PR #6, opened 2025-05-05
business idea generator

KaushikML:
All models generate business ideas according to the ongoing trends , judge tells how likely it can succeed


ed-donner:
nice @KaushikML ! How great to have you in this repo too! Actually we make a business idea generator on the course in later weeks.. you're several steps ahead of me..

PR #12, opened 2025-05-06
Adding graph visualization

sriggelsen:
Visualizing agent graphs really helps me understand and validate logic so suggest adding it here


ed-donner:
Hey @sriggelsen this is fantastic - thank you so much for this!
I have 2 requests:
- please could you move this into community_contributions folder for me.. thanks so much
- would it be possible to use `uv add` instead of `uv pip install`, or was there a reason that you went the pip route?

PR #14, opened 2025-05-07
ADDED project in 2_openai/community_contributions with name customer_care_agents

saadrasheeddev:
This project is built using Google ADK. 

This project contains 4 agents and a simple tool that can query a dataframe and generate answers straight from live databases. 
Agents receives query, understands it, rewrite query in database language for fellow agents to generate actual query. 
a query generator agent generates a python pandas query for now for simplicity and is given an output schema of pydantic model.
another agent then takes this query and executes it using the tool
ADK in Action


ed-donner:
hey @saadrasheeddev this is so terrific - huge thanks again for this. Hey I have 2 minor requests:
- please could you put this into a separate subdirectory within the week2 community_contributions folder - that way people will see this as an isolated project and learn from it
- please could you take pyproject.toml and uv.lock out of the PR, and instead add a comment in a README in your folder that students should run `uv add ...` to set things up - otherwise we have a merge conflict with uv.lock which could be hard to resolve..

ed-donner:
Absolute masterpiece @saadrasheeddev - thank you so so much

PR #15, opened 2025-05-07
Upgraded deep research

KaushikML:
added clarifying agent , manager agent


ed-donner:
wow this is incredible @KaushikML thank you so much

PR #17, opened 2025-05-08
Free CrewAI Agent

palbha:
Notebook to create Financial Researcher CrewAI Agent with togetherai free api key


ed-donner:
hey @palbha thanks so much for this. A couple of requests:
- Would you be OK to clear the outputs? It makes it easier to see what's going on, and is more efficient in the repo
- I notice that you do some "pip installs" - when working with uv, it's better to use "uv add" instead - did you look into that approach?
Thanks so much
Ed

palbha:
Hi @ed-donner, Thanks for your comments

1.   Sure I would remove the outputs  & update 
2.  For this I am not using uv but just tried to run through the google colab hence the above,but let me know if you would like me to remove this assuming the users using the repo would be using uv instead

ed-donner:
Actually I think it's great the way you did this @palbha - a really useful example for others, not using uv. Adding now with great thanks!

PR #18, opened 2025-05-08
Adding graph-visualization block to openai lab 2

sriggelsen:
Adding graph visualization to OpenAI lab 2 notebook. Only change to original notebook is the cell beginning with '# Visualize the graph' comment.

Note: Running `uv add "openai-agents[viz]"` will update `pyproject.toml` and `uv.lock` with the viz extension. If adding this to the main lab2, consider update these files as well to let new students automatically get the viz extension when setting up their venv.


ed-donner:
this is excellent @sriggelsen thank you so much! A really valuable addition.

PR #22, opened 2025-05-09
Exercise on  lab2 notebook using synthesizer/improver pattern

mallocode300:
Hi Ed, I have slightly modified the agentic pattern on lab2, to switch from judge from a synthesizer.


ed-donner:
Terrific @mallocode300 thank you very much - merged with thanks

PR #23, opened 2025-05-09
Book_Researcher_AI

KaushikML:
New Book researcher crew ai project which finds trending books of particular genre , researches the genre to give details and finds top novelist of that genre


ed-donner:
Brilliant @KaushikML - cool idea, nicely executed! Thank you as always,

PR #26, opened 2025-05-11
:refactor: :recyle: - Removing the google genai library from the..

ChorjoDev:
…project and provided installation instructions instead


ed-donner:
hey @ChorjoDev can you give me more background on this.. what's the reason for your suggestion.. does this help with the setup process? Actually I can't think why we have a dependency on this package at all.. maybe I can just remove it?

ChorjoDev:
Hey @ed-donner, sorry yes of course. 

It was based on a comment on my previous PR where I added lab 1 but using Gemini. You asked me to remove the library from the main repository and to provide installation steps instead to install it

ed-donner:
oh gosh I remember now, sorry, of course. Thanks so much for this. And thanks for replying so fast!

PR #27, opened 2025-05-11
Updated week 2 deep research

pKap4:
Added a contextualizer to get the clarifying answers from the user and generate an updated query before passing it to the research manager


ed-donner:
Thanks so much @pKap4 !
I've discovered btw that there was a nasty bug in my code that I think you've carried in to this too. In Pydantic, if you use the approach for using docstring comments like this, it actually won't be provided to the LLM as a description of the field. You'd need to use the Pydantic Field approach. I've fixed this in my code - no rush, but sometime you might want to fix yours too. Interestingly everything works anyway because LLMs can figure it out from the name of the variable!
Ed

PR #33, opened 2025-05-14
Add example of using lifecycle hooks to monitor openai agent activity

dalemartyn:
### Agent logging with lifecycle hooks

This PR demonstrates using OpenAI's Agent lifecycle hooks to print the steps the agent is taking to solve the task.

Based on the [agent lifecycle example](https://github.com/openai/openai-agents-python/blob/main/examples/basic/agent_lifecycle_example.py) in the openai agents examples.

This gives us some idea of what is happening without having to use the traces screen.

It also demonstrates using a class to instantiate multiple MCP servers from multiple configs, which makes using several MCP servers at once more ergonomic in a notebook.


ed-donner:
That's really great @dalemartyn thank you so much. The MCP Server manager is very elegant. Thanks for this valuable contribution.

PR #35, opened 2025-05-14
updating prompt

josdijkstraco:
Updating the prompt to make it more robust, i.e. don't act on anything not related to the resume, summary, etc.


ed-donner:
Hey @josdijkstraco this seems like an excellent change. Did you consider adding this instruction to the system prompt, rather than adding it to the user prompt like this? It might seem a bit confusing to have some user prompts being questions from the user, and the most recent being instructive in this way?

PR #39, opened 2025-05-17
Sidekick with calender integration

KaushikML:
Sidekick with calender where you can check upcoming events and schedule events(Although its very hectic to set up google app , api, json its very time consuming i have included how to do it in readme file.)


ed-donner:
Awesome @KaushikML thank you so much.
Ed

PR #42, opened 2025-05-18
Add deep_research_with_clarifying_question

nkthien:
Update the orginal deep research agent (week2-day5) with capacity to generate clarifying questions based on the topic and send the answers and the original query to ResearchManager to conduct the final report.


ed-donner:
Thanks so much @nkthien - great project and very valuable..

PR #49, opened 2025-05-26
Six Thinking Hats simulator for 1_foudations_lab2

elixirgam:
Simulating Six Thinking Hats analysis technique


ed-donner:
That is inspired! Brilliant contribution @elixirgam , thank you

PR #52, opened 2025-05-28
Updated notebook 2_lab2.ipynb to let each model judge others

dynologic:
Making the changes so that each of the 6 models can judge each others responses and rank them. I have a server running Ollama as well. Store those results in a list and then do a weighted average to truly come up with the best response.

The code to determine the averages was aided by Cline in Cursor and Claude 4 as well.

Side Note: I have run this 4 times and interestingly, Gemini 2 seems to come out the winner every time.


ed-donner:
That's an excellent change @dynologic and fascinating about Gemini 2..
Thanks so much for sharing this great contribution.

PR #53, opened 2025-05-29
Added example notebook using deepseek and mlflow for tracing 

akhilmakes:
Example notebook using deepseek as the LLM and using mlflow for tracing the output.


ed-donner:
That's great @akhilmakes - really good to see mlflow for tracing. Thank you.

PR #55, opened 2025-05-31
Lab 2 mcp work around on windows

Bensk-96:
Hi Ed,

I found a workaround for running MCP in Lab 2 of Autogen in windows environment. I created a separate `.py` script for the MCP-related code to avoid the `fileno` error seen in Jupyter notebooks.

The workflow is as follows:
1. Start the MCP server in a separate terminal using: `uvx mcp-server-fetch`
2. Then, in another terminal, run the MCP script using: `uv run mcp_fetch.py`

The LLM’s output will be printed directly in the terminal.

Along with the `.py` script, I’ve also added a `README.md` with clear instructions for this workaround. Hopefully, this helps other students facing the same issue.

Best,  
Benjamin


ed-donner:
Gosh thank you so much for this @Bensk-96 - super clearly explained.. great solution and thorough docs. Massively appreciated. You've added value for everyone...

PR #56, opened 2025-05-31
Deep research v2

Bensk-96:
Hi Ed,

I've made some modifications to the Deep Research app in the OpenAI Agent SDK section. This updated version — Deep Research v2 — includes the following enhancements:

1. The agent asks up to 3 clarification questions when needed.
2. Added a toggle to control whether an email is sent or not.
3. Allows selection of research depth (5, 10, or 20 search results).

Could you please take a look? I hope you and other students find it useful and interesting!

Best regards,  
Benjamin


ed-donner:
It's terrific @Bensk-96 thank you so much.

I should mention that another student discovered a horrible bug in my code (that in my defense came straight from OpenAI code). I was using a shortcut for Pydantic like:

```python
short_summary: str
    """A short 2-3 sentence summary of the findings."""
```

But in fact that won't work - the description won't be sent to the Agent. You'd need to do it the old-school way:

```python
short_summary: str = Field(description="A short 2-3 sentence summary of the findings.")
```

i've made that update in the main project.. you'll probably find you get better outcomes if you make similar updates..

PR #57, opened 2025-06-01
Lab1

adithya-ananth:
I've used Ollama locally to run the code using the Llama3.2:1b model. I hope this helps those who desire to use models without spending on API keys.


ed-donner:
Thanks so much @adithya-ananth - this will be most helpful for others.

PR #61, opened 2025-06-02
Module 1 - Labs 1, 2, and 3 implemented using the OpenRouter API

rodrigo-amc:
Using the OpenRouter API allows us to use a single API key to access a lot of LLM models, including free models.


ed-donner:
Hey @rodrigo-amc - very helpful thank you. It looks like you may have accidentally included your own files in the main "me" directory in week1. Is that right? Would you be OK to remove them before I merge.. thanks so much

rodrigo-amc:
Hi @ed-donner , thanks for the feedback. I've made the requested changes.

ed-donner:
Thanks so much - terrific - this is an excellent contribution

PR #65, opened 2025-06-07
InsightFace comparison using URLs as an MCP server.

johnbikes:
I tried slimming the dependencies down just to what is extra here. Let me know if this needs to change in any way. I'm happy to remove the code headers as well! Thanks!


ed-donner:
Wow this is so great @johnbikes - a really substantive PR. Thank you for sharing this!

PR #72, opened 2025-06-13
Add notebook for section 1: llm_requirements_generator.ipynb

maldaurore:
This notebook generates and validates functional and non-functional software requirements from a natural language description, and classifies them using the MoSCoW prioritization technique. It also ensures that the user description has all necessary information before continuing to the next steps.


ed-donner:
An interesting application @maldaurore thank you

PR #78, opened 2025-06-16
Implementing a Routing Workflow based on Problem Complexity

ggsiqueira:
In the original Lab 2, we explored the Orchestrator–Worker pattern, where a planner sent the same question to multiple agents, and a judge assessed their responses to evaluate agent intelligence.

This notebook extends that design by introducing multiple judges for more robust evaluation and a routing component that classifies task complexity. Based on the aggregated rankings, the router directs simpler tasks to smaller models and reserves more complex ones for advanced LLMs, optimizing resource usage across the system.


ed-donner:
Very interesting @ggsiqueira - thank you for working on this and sharing your results

PR #79, opened 2025-06-17
Enabling google search tool for deep research

subbu-sw:
The deep research Agent is very useful to research and summarise information from the internet. Here is a pull request that utilises google search instead of openai's hosted tool to search for the info. This reduces cost significantly and you can now use this tool for personal summaries and research. 

I have used the python sample from google custom search JSON API (details [here](https://developers.google.com/custom-search/v1/libraries)) to build this tool. You will need to create a google search API key and search context and add it to the .env file before starting to use the google custom search JSON API.

```
GOOGLE_SEARCH_API_KEY=
GOOGLE_SEARCH_CONTEXT=
```

With this tool, depending on your google search API usage quota, you should get to do upto 100 free searches a day.

### DIFF
--- a/2_openai/community_contributions/deep_research_using_google_search/README.md
+++ b/2_openai/community_contributions/deep_research_using_google_search/README.md
@@ -1,0 +1,6 @@
+---
+title: deep_research using google_search
+app_file: deep_research.py
+sdk: gradio
+sdk_version: 5.31.0
+---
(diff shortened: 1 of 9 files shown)


ed-donner:
Excellent - thank you so much @subbu-sw - very useful to have a google search example

PR #84, opened 2025-06-20
Add security design review agent notebook

abathgate007:
Added security_design_review_agent.ipynb which takes requirements and design for OWASP Juice shop and then calls several LLM models to perform design reviews, then it ranks them, then it merges them into one report with all the findings from all the models.


ed-donner:
Intriguing @abathgate007 - thank you for this!

PR #85, opened 2025-06-21
slack as alternative in foundations 1

dexterchan:
Slack is free at any time at this moment. It is a battle tested notification app.

See if it works for you.

### DIFF
--- a/1_foundations/4_lab4_slack.ipynb
+++ b/1_foundations/4_lab4_slack.ipynb
@@ -1,0 +1,467 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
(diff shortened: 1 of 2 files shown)


ed-donner:
Hey @dexterchan - this is great, but would you be able to make the change into the community_contributions folder? Also include a note that people should `uv add slack_sdk` but don't change the project dependencies if possible - thanks so much

dexterchan:
@ed-donner : done.... I changed the file to community contribution folder. reverted requirement.txt. Also, the notebook contain the line 
uv pip install slack_sdk==3.35.0

ed-donner:
that's great! Thanks so much for this terrific contribution..

PR #97, opened 2025-06-30
Labs in Week1

tlconde:
Created a multi-modal product selector with specific user input. 

Created a career conversation with faq document for "safe" fallback and to limit certain information provided. Attempted embedding RAG to make this happen.


ed-donner:
Wow that's a pretty hefty project @tlconde - thank you so much for sharing and contributing.
Ed

PR #98, opened 2025-06-30
Language translation evaluator-optimizer

shandran:
Here is the evaluator-optimizer model that takes public-domain works (in this case George MacDonald's The Princess and the Goblin), and performs translation using Gemini model, evaluates the translation quality with an OpenAI model and has some (non-perfect) error correction and backup translation with a Claude model.


ed-donner:
Thanks so much @shandran, a valuable project. Hey - it looks like you might be changing files in the main directory here; would you be able to update with all changes in the community_contributions folder. Also double check that outputs are cleared if poss. Thanks so much
Ed

ed-donner:
Hey @shandran thanks for the updates.. it's unfortunately still looking like this is changing a number of files in the main repo..

PR #100, opened 2025-07-02
Added folder for section 2: agent_manager_refactor

maldaurore:
Refactored the project so:

- Each research stage is implemented as a standalone tool.
- Generates clarifying questions.
- A **manager agent** controls the flow and selects tools dynamically.
- User interacts with the manager, enabling a **more conversational** research experience.
- Rather than hardcoding the user's email address in the script, the agent dynamically prompts the user for their address and uses SendGrid to send the report to that input


ed-donner:
Excellent improvements, @maldaurore , thank you so much!

PR #106, opened 2025-07-05
Add evaluator-optimizer pattern to 2_lab2: llm-text-optimizer.ipynb

murugan-aravind-rxg:
- Added evaluator optimizer pattern to lab2.  
- Used Groq2 llm as a text optimizer. 
- Used OpenAI as an evaluator. 
- File Name: llm-text-optimizer.ipynb


ed-donner:
Great @murugan-aravind-rxg thank you!

PR #119, opened 2025-07-11
week2_lab3_exercise

bluebells1:
Update to week2 (openai) lab3 to include:
- Use of Claude ("claude-3-5-sonnet-20240620") as one of the models
- An email class for consistent structure
- Inclusion of standard footer and log of all emails sent
- I also used google to send the mails rather than sendgrid with my .env file as a lazy way to include to and from emails and my app password.

I ran this a few times to test it and I found it interesting that Claude came back a couple of times complaining about being the Head of Compliance, e.g. "As an AI assistant, I don't actually hold the position of Head of Compliance at ComplAI. I'm here to help you craft messages, not to impersonate real people or positions".  I thought this was interesting!


ed-donner:
Hey @bluebells1 - once again, a valuable contribution from you. Yes that is interesting about Claude - I guess it ties to Anthropic's strong safety and alignment culture. I wonder if you can correct for that through the system prompt / instructions..

PR #121, opened 2025-07-11
completed day one activity

abali2509:


abali2509:
finished first days exercise

ed-donner:
Hey @abali2509 that's great! Would you be OK to move it into community_contributions? I can't easily change files in the main repo... thanks so much
Ed

PR #123, opened 2025-07-13
Added my contributions to community-contributions. Re: ollama llm in agents.yaml

afrankel:
Within 3_crew/coder - Added additional backstory prompts in the case when using the ollama llm. The additional prompts are needed since the llm is looking for the python libraries_used parameter.

Without the additional prompt the execution fails due to an error message asking the user to add the libraries_used parameter. It's expecting an array [].


ed-donner:
Brilliant @afrankel - thank you so much, and thanks for going through the painful PR process! Much appreciated

PR #130, opened 2025-07-16
Adding chatbot with RAG and Evaluation

hamurda:
A chatbot app that uses LangChain RAG for chunk retrieval, OpenAI for generation, and Gemini for response evaluation. It also records user emails to Google Sheets or CSV fallback


ed-donner:
terrific @hamurda  - that's a substantive addition, thank you!

PR #138, opened 2025-07-18
use current date for tasks

esteinholtz-cloudera:
I had problems getting a fully up-to-date report using the notebook supplied, so I added usage of current date.


ed-donner:
hey @esteinholtz-cloudera - any chance that this could go into community contributions?
If you'd like to make a change to the main repo then I do agree that this is a good change to make, but I'd be fussy and ask that you come up with the current date in a more expected way - just import date at the top of that module, and set the current_date based on that.. or is there a particular reason for the dynamic import?

PR #142, opened 2025-07-19
Files for a simple AI agent app.

glafrance:
simple-tools-usage is a very basic example of using the OpenAI API with a tool.

The "tool" is simply a Python function that:
- reverses the input string
- converts all letters to lowercase
- capitalizes the first letter of each reversed word


ed-donner:
Very nice, @glafrance - a simple and clear example. Much appreciated.

PR #147, opened 2025-07-21
feat/2_lab2_exercise : new reflection pattern added

jessicakuijer:


ed-donner:
Terrific @jessicakuijer thank you for sharing

PR #149, opened 2025-07-22
langgraph_lab4_consenthandling

bluebells1:
Update to the langgraph lab4:
- Includes handling for cookie pop ups (both dialogs and html).  Unlikely to work 100% of the time as depends on button naming.
- Includes the openai reasoning mini model (o4-mini) as the worker.  I found that gpt-4.1-mini (or gpt-40-mini) struggled with the right searches to use in some cases.
- Updated the worker prompt to push for tool usage even if answer available in training data.
- Includes Claude as the evaluator agent.
- Update to the evaluator agent system prompt as Claude was being very pedantic and rejecting everything!


ed-donner:
Thank you for contributing this @bluebells1 - I'm sure a lot of people (including me!) will find this most useful

PR #154, opened 2025-07-23
Add Kyra's lab 4 solution (cleared outputs)

kyra-nank:


ed-donner:
Thanks so much @kyra-nank - thanks for building & sharing

PR #155, opened 2025-07-24
Sam's Deep research agents including Clarifier, Evaluator and Optimiser agents

samsriram712:
Have extended the deep research application by adding clarifying questions to the original query and also evaluator and design patterns using the OpenAI endpoints for Gemini and Groq. Used vibe coding with chatGPT to develop the solution. Tried Grok3 but the steps and code snippets were more complicated than chatGPT (maybe Grok4 might be an improvement). Perplexity was way off the mark as a vibe coding tool (at least for a novice like me).


ed-donner:
Well that's great - and congrats on the result! Thank you @samsriram712

PR #158, opened 2025-07-26
Deep research refactored

hamlet0808:
This is a my version of deer research agent, using Openai Agents with gpt-4o-mini, 
also added FastAPI app for the agent.

Please share with me your feedback, which will be highly valuable for me.

Best, Hamlet


ed-donner:
Very nice @hamlet0808 - thanks so much for that - great to see a proper implementation with FastAPI

PR #159, opened 2025-07-26
Sams/deep research update

samsriram712:
Updated the UI to accept the clarifications as textboxes instead of checkboxes. Also removed the send_email function as I was getting emails when others were trying it out from Hugging face. The output shows only on the browser. A few other UI enhancements to show progress status when the full research is running.


ed-donner:
Thanks so much @samsriram712 - valuable updates

PR #163, opened 2025-07-28
App to generate a novel.

glafrance:
An agentic AI application to generate a novel, after prompting the user for specifications for the novel.


ed-donner:
That's a terrific idea and nicely built! Thank you for sharing, @glafrance

PR #164, opened 2025-07-28
Handling Real-Time Event Dates in Interactive CVs

fmezacr:
Solution for that issue


ed-donner:
Very nice - thank you @fmezacr - and thanks for sharing

PR #166, opened 2025-07-28
Lab1 contirbution using gemini

aayush2095:
Please only consider community contributions


ed-donner:
Hey @aayush2095 - thanks for that. Unfortunately I don't think I can "selectively merge" - I think you'd need to update the PR to not include the files which you don't want to change. If you're not sure how to do that, you could also close this and create a new PR. Thanks so much

PR #168, opened 2025-07-28
Community

adityamandlekar:


iankisali:
Hi @adityamandlekar. These are great contributions. However, all contributions are made through the community_contributions folder. For example, your contributions should be in 1_foundations/community_contributions folder, same as other students.
Your ipynb files seems to be large, probably caused by pushing the outputs. It would be better to clear the outputs before pushing.
Also I've noticed that you've pushed a certificate file. This might be a security issue.

adityamandlekar:
Opening a new one

PR #186, opened 2025-08-03
Add Sergio's lab 1 solution with Azure OpenAI

seryirobles:
Adding Azure OpenAI option. Will need an Azure OpenAI resource with the models that are being used in the course deployed beforehand


ed-donner:
Great! Thanks so much @seryirobles - thanks for sharing.

PR #187, opened 2025-08-05
Created my solution to the exercise for week 2

Myde001:
Added the clarification question and converted agents to tools.


ed-donner:
very nice indeed @Myde001 - thank you for sharing

PR #188, opened 2025-08-05
Update execution counts and outputs in lab1_gemini notebook; enhance example questions and responses

dipayanrana:


ed-donner:
hey @dipayanrana so generally I prefer to have labs without outputs saved, because it's cleaner in git. Was there a particular reason that you wanted to update this lab? Thanks

PR #201, opened 2025-08-10
2 lab2

primevalsoup:
This notebook extends the original by adding a reviewer pattern to evaluate the impact on model performance and give the model an opportunity to revise its original answer based on feedback.  It compares impact of the revision process on each models ranking when only that model is given the opportunity to revise based on the feeback.


ed-donner:
Thanks for sharing your work! A most valuable contribution.
Ed

PR #212, opened 2025-08-12
Add Spotfy tool

primevalsoup:
**Adding a Spotify Tool - Musically You!**

This version of the notebook introduces a Spotify tool that can query your listening history from Spotify to extend the domain of questions the chatbot can answer to include your musical tastes.

Unfortunately, it's a bit of PITA to get acess and refresh tokens for Spotify.  The process requires connecting to an authentication end point while logged in to Spotify and then processing a callback.  To make this easier, instructions along with a small app that can be deployed to HuggingFace Spaces using Gradio are included at the end of this notebook.


ed-donner:
Such a great idea.. Thanks for sharing this great addition.
Ed

PR #219, opened 2025-08-15
Update SETUP-mac.md

hashi-rich:
Alternative crewai install method


ed-donner:
Hmm thanks for sharing.. are you sure it's ok for me to hardcode version 0.121.1 like that? Might this mean that in 6 months time people might be installing outdated versions of the crew tool?

PR #220, opened 2025-08-15
Add Steve's deep_research_with guardrail

stevehome:
Added output guardrail on report length.
Not actually activating tripwire, but using flag to show report or not.


ed-donner:
Greatt.. Thanks for sharing your work! A most valuable contribution.
Ed

PR #221, opened 2025-08-15
feat/add pushover notifications instead of sendgrid emails

jessicakuijer:
Hi Ed Donner!
I've added pushover calls over sendgrid to follow the course because I couldn't use sendgrid (forbidden email sender authentification and it seems complicated for sendgrid to unblock my email address).

it works correctly, let me know if this can be an accepted PR.


ed-donner:
Terrific - hugely appreciated - many people ask for this. Thanks for sharing your work! A most valuable contribution.
Ed

PR #224, opened 2025-08-16
feat/add pushover notifications instead of sendgrid

jessicakuijer:
Hoi Ed Donner,
Last but not least, I've replaced here too sendgrid by pushover notifications.
Thanks again!
Jessica


ed-donner:
Well thank you again!!
Ed

PR #231, opened 2025-08-19
database backend schema builder agent

Excergic:
I have build Ai agents that can understand business requirements, build DB schemas with SQL query and also reviews the query all using CrewAI.


ed-donner:
That's a very substantive project @Excergic - thank you for sharing - great result
Ed

PR #234, opened 2025-08-21
Add financial_researcher_guard derived from financial_researcher

stevehome:
Added 2 guardrails in crew.py .


ed-donner:
Very valuable @stevehome - thank you for sharing

PR #235, opened 2025-08-21
Added extra line in judge's prompt

CleverCoder03:
I followed every steps perfectly while running 2_lab2.ipynb, though i encountered an error. 

The error was something related to the _results_ from the judge's prompt
`ValueError: invalid literal for int() with base 10: 'competitor 1'`

This was caused beacuse of the output i recieved was:
`{"results": ["competitor 1", "competitor 2", "competitor 3", "competitor 4"]}`

So i just added extra line which termed:
`Answer only the number for example
{{"results": ["1", "2", "3", ...]}}`
In the judge's prompt and it fixed my issue.

And then when i ran my code, it worked perfectly fine.

### DIFF
--- a/1_foundations/community_contributions/2_lab2-judge-prompt-changed.ipynb
+++ b/1_foundations/community_contributions/2_lab2-judge-prompt-changed.ipynb
@@ -1,0 +1,476 @@
+{
+ "cells": [
+  {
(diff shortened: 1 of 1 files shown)


ed-donner:
Gotcha - thank you @CleverCoder03 - nicely found and fixed

PR #240, opened 2025-08-25
Add Sanjeev-Wahi's update to use default current date, every time you…

opensourcefintechinc:
… run the app to research the most recent data online. Additionally, by adding the current date at both mainy.py & in the tasks.yaml file, agents do a better job with an extended report: Week3-CrewAI Multi-Agent Financial-Researcher


ed-donner:
Great - many thanks @opensourcefintechinc

PR #244, opened 2025-08-27
Finance copilot

gaurangpatel79:


ed-donner:
Hey @gaurangpatel79 - I really appreciate this fantastic project. With more than 60 files and 13,000 lines of code, this is more of a major initiative than a mere contribution.
Can I suggest a different way: perhaps you could submit a single notebook that describes your project with a screenshot, and then have a URL link to this as a repo in your github? It seems like that would work better - this should be it's own repo, rather than a directory in a contributions folder..
Let me know what you think - thanks
Ed

PR #246, opened 2025-08-27
feat: implement orchestrator-workers workflow demo for lab 2

lachilles:
add orchestrator-workers workflow demo contrasting lab 2's evaluator-optimizer pattern


ed-donner:
Ed

PR #266, opened 2025-09-04
added a Go MCP server to run locally which uses a local Python Text E…

stevef1uk:
As discussed a long time ago here is an extensible Go MCP server (originally based off the one from Prasanth Janardhanan but now so radically changed and restructured it really isn't his anymore. The new author is Cursor ;-)

The Go MCP server has been configured to use one Tool and that is a Python API which takes an OpenAPI type request and returns the sentiment of the text using a standard Huggingface dataset: https://huggingface.co/boltuix/bert-emotion

All code can be run locally on Linux and Macs. I haven't tested on Windows as I don't have any Windows machines.

The code was developed on my big Linux server with a nice GPU but it was also run and tested on a RPi 5.


ed-donner:
Wow this is amazing - a huge contribution @stevef1uk thank you

stevef1uk:
Thanks Ed. I have used this and the fine-trained model Gemma3:120 in a docker container and all described here: https://emotion-service-api.sjfisher.com/

PR #277, opened 2025-09-07
Fixed TLS cert issue for self

cmohite:
Using work laptop caused SSL errors:
[SSL: CERTIFICATE_VERIFY_FAILED] unable to get issuer certificate (_ssl.c:1010).

Replaced default OpenAI client initialization with an httpx.Client that uses
the custom bundle from SSL_CERT_FILE, which resolves the connectivity issue.


ed-donner:
Thanks! Also see Q15 here for the more permanent fix possibilities: https://edwarddonner.com/faq/

PR #290, opened 2025-09-14
Week1 - Linkedin Profile Chat, with Pushover and Sql Db Check

Harshbhama:
Features - 

1. Use gradio to chat with openAI API, by providing Linkedin profile, along with resume pdf.
2. If answers are not found, send pushover notification.
3. Now for answers not found case - 
  3.1 Hit Database to check for the Predefined Questions and Answers. 
  3.2 Again call openAI api call, to match the relevant answers. We are consuming all Sql Table data into openAI, then asking
OpenAI to find relevant answers.
4. Provide  those answers to the gradio chat.


ed-donner:
Ed

PR #295, opened 2025-09-17
Google ADK Sales Agent

dhanush-r-m:
By using the gemini models


ed-donner:
Terrific! Many thanks - always great to get an ADK example in there.

PR #296, opened 2025-09-17
Add trading_with_short

stevehome:
This is based on the trading example in the course. It has a simplistic approach of allowing short selling up to a financial value of MIN_SHORT_BALANCE per equity. There should really be an overall limit for all equities.


ed-donner:
That's great! I've been looking forward to someone doing this..

PR #298, opened 2025-09-21
Akhil changes to make the WF with handoffs

akhilmanidara-maker:
HI Ed,

I tried my shot at converting week 2 to introduce Handoffs instead of using functions. 
however, when i run the code, only first agent -> Planner agent gets triggered. and nothign happens after that. 

I tries a bunch of other things using chatgpt and Kiro help. However, i was not successful in acheiving the outcome. 

Could you help me understand how can i 

1. introduce multiple handoffs through orchestration model and 
2. hoe can invoke a for loop to trigger multile sear agent handoffs in parallel?


ed-donner:
Thanks for the contribution! I'm not sure I understand the question.. you can give every agent a handoff, and it will use it if appropriate.. maybe post the question on udemy or by email? Thanks much

PR #315, opened 2025-09-29
OpenAI Agents SDK: Deep Research Example with OpenTelemetry Tracing

dmitchsplunk:
Hi Ed, per your suggestion on LinkedIn, I've created an example that adds OpenTelemetry tracing to the OpenAI Agents SDK Deep Research example shared in week 2, day 4 of the course.  The example uses Jaeger, which is open source, so anyone can use it. 

Note that I included an updated version of pyproject.toml and uv.lock, as I found it necessary to modify the version of the openai-agents package to be compatible with the openlit package I used to capture the trace details.


ed-donner:
Makes sense - thanks so much, and thanks for explaining that. And, great work! And very nice to see OpenTelemetry here.

PR #318, opened 2025-09-30
feature(handover instructions): tweaked instructions for clearer handoff

paulmboyce:
tweaked instructions for clearer handoff by:
- adding a 'Your task completes when ...' as the final statement. 
Tested on gpt5 nano (which being small needs more strict instructions)


ed-donner:
Yes - nicely experimented and solved - thank you!

PR #320, opened 2025-10-03
Add Indira_1_lab1 (cleared outputs) to community_contributions

IndiraV5:
Added the solution of lab_1 of the 1_foundations.


ed-donner:
Great! Thanks for making this valuable contribution and sharing your work with us all.
Ed

PR #323, opened 2025-10-05
Sales manager having multiple recepients

ainanihsan:
Modified the original code to have the email manager send emails to multiple recepients. Placeholders for {first_name} and {company} are added to send directed emails.


ed-donner:
Ed

PR #327, opened 2025-10-07
Feature travel planner chat agent

tanvimalwe:
- This application utilizes the **Gemini API** to function as a sophisticated travel planner.
- Takes detailed traveler information from travel_summary.txt and generating a comprehensive, strictly-formatted trip itinerary.
- The user interface is built using Gradio, providing a convenient chat environment.
- The final itinerary which the user is happy with, can be saved directly to a file via the model's tool-calling capability.


ed-donner:
Great! Thanks for making this valuable contribution and sharing your work with us all.
Ed

PR #338, opened 2025-10-13
Update CrewAI memory configuration per latest docs (verified Oct 2025)

emads22:
- Updated memory configuration to use `memory=True` instead of manual instantiation
- Added embedder parameter for OpenAI embeddings
- Fixed CHROMA_OPENAI_API_KEY issue for local Chroma
- Aligned with the latest CrewAI documentation (verified October 2025)
- Tested and verified with log output and Pushover notification

Note: This reflects the new Crew approach and differs from the video example.


ed-donner:
Oh my goodness thank you so much yet again.. brilliant

PR #339, opened 2025-10-13
Deep research ms

smudger007:
My attempt at the deep_research openai challenge. 
Put in some additional gradio features. 
Added in a clarifier agent.
Created a manager agent, which uses the other agents as tools. 
A little slow but does the job :-)


ed-donner:
Wonderful - thanks so very much!

PR #343, opened 2025-10-15
week_1(Yasaman_Forouzesh)

YasamanForouzesh:
Can I also get feedback on file structure as well? I'm coming from Go (Golang), where there's a standard structure for creating files. I'm not very familiar with how this works in Python. How should Python files be organized?

I have repo on my github that contains whole file like uv and env files here I just put these three files that have all my codes.


ed-donner:
Hey! It looks great to me. Python is unopinionated about how you organize your code into modules and packages. It's more important to focus on the logical separation of concerns and a good design than the file structure. And this looks great!

PR #346, opened 2025-10-20
honeypolygon/lab1 solution

honeypolygon:
### Hello Donner!
AI Agent systems are fascinating — I actually rushed through the Python basics just so I could understand this part of the course! 😄  

---

### Concept
This notebook explores a structured AI debate framework.

When a decision or opinion needs to be made, two LLMs take opposing sides:
- **Pro side** argues in favor of the topic.
- **Con side** argues against it.

Each argument is then **independently verified and revised** by another LLM acting as a verifier, improving logic and factual consistency.  

Originally, I also planned a more advanced structure — where the verifier would **send the improved argument back to the original debater LLM**, and only if it passes verification again would it be forwarded to the **Judge LLM**.  
That feedback loop would simulate an iterative refinement process, but it turned out to be quite complex to coordinate, so this version represents the simplified foundation of that idea.

Finally, a **Judge LLM** evaluates both refined arguments and determines **a clear winner**, with no ambiguity allowed — one position must be adopted as more valid overall.

The goal is to simulate an AI-based deliberation process that mirrors balanced reasoning, self-correction, and decisive judgment.

---

### Background
I'm a beginner in Python and still learning through the course materials.  
This notebook was created as part of my learning process — it may share some structure or ideas with the lecture content, but I developed it independently to understand multi-agent reasoning and coordination.  
Please kindly excuse any overlap with course examples.

---

### Notes
- All outputs have been cleared.  
- The file is safely located under `community_contributions/honeypolygon/`.

### DIFF
--- a/1_foundations/1_lab1.ipynb
+++ b/1_foundations/1_lab1.ipynb
@@ -1,0 +1,395 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Welcome to the start of your adventure in Agentic AI"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "<table style=\"margin: 0; text-align: left; width:100%\">\n",
+    "    <tr>\n",
+    "        <td style=\"width: 150px; height: 150px; vertical-align: middle;\">\n",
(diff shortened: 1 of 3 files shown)


ed-donner:
This is great! Thanks so much for sharing. One issue - this PR seems to include files outside the community contributions folders. Let me know if this was intentional - otherwise please move all changes within community contributions when you get a chance. Thanks so much
Ed

PR #347, opened 2025-10-20
Weather agent using MCP

gabormeresz:
Implementing a weather MCP solution based on week 6 day 2 examples


ed-donner:
A classic! Thank you so much.

PR #349, opened 2025-10-21
Annpa's Lab2 exercise solution-Added Evaluator optimizer workflow

AnnpaS18:
Added Evaluator optimizer to make the LLM guess the question, that initially gave the answer, added a max retry limit.


ed-donner:
Great - thank you so much

PR #354, opened 2025-10-26
Feature/deepseek lab2 integration

MoizMujtaba:
Agentic workflow to generate a complex question, have more than one LLM respond back to the question and then stack rank the participant responses based on LLM evaluation.


MoizMujtaba (commented):
Adding my name in signature.

MoizMujtaba (commented):
approved

PR #355, opened 2025-10-26
Update crew.py to update fieldname for latest version of CrewAI

ackmase:
Needed to update the fieldname from "model" -> "model_name" to ensure compatibility with the latest version of CrewAI.

### DIFF
--- a/3_crew/stock_picker/src/stock_picker/crew.py
+++ b/3_crew/stock_picker/src/stock_picker/crew.py
@@ -103,7 +103,7 @@ class StockPicker():
                         embedder_config={
                             "provider": "openai",
                             "config": {
-                                "model": 'text-embedding-3-small'
+                                "model_name": 'text-embedding-3-small'
                             }
                         },
                         type="short_term",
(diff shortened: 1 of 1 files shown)


ed-donner:
Oh jeez.. this is tricky @ackmase .. did you recreate this Crew project and then run it, or did you run it as-is from the repo? I think that the uv.lock means that it will expect the old way if you run it direct from the project..? Or is that how you did it..
This CrewAI upgrade has been quite problematic!
Ed

ackmase:
Oh boy, that's a really good point. I don't actually remember if I ran as is in the repo or if I recreated the project. I think your theory that the uv.lock would freeze it on the old version makes sense, so I probably recreated it. Please feel free to disregard this pull request, and sorry for bugging you! 

(I just finished your full loop of 4 LLM courses. Thank you so much! I feel like a completely new person, and I'm ready to take on the future of AI :))

ed-donner:
gotcha ok - I need to figure out how I deal with the Crew update..
Oh that's amazing about all 4 courses - thank you!!

ed-donner:
and congratulations! And what a time to be diving into this field.. crazy..

PR #360, opened 2025-10-30
BDD Agent

sngo:
Implement BDD Agent using playwright , file system MCP server and custom build assertion mcp server


ed-donner:
A big project! Many thanks @sngo for building & sharing. I'm excited to try it out!

PR #363, opened 2025-10-30
First 3 weeks contributions

Shmacked:
Week 1 2_lab2 - It uses each LLM to judge the output from the other LLMs.
Week 1 app.py update - just added pathlib.Path to check for summary.txt and linkedin.pdf, and made it into a class.
Week 2 2_lab2 - I updated it to use the SerperAPI and IPInfo APIs (described in the file) to get your local news, create summaries, pick the best summary, and email it to you. I wanted it to send the image attached to the article, but there are some security vulnerabilities with that, that I couldn't work past.
Week 3 stockpicker - I updated a few files. I had a hard time getting the LLM to respond with just a "message", so the push notification failed multiple times. I added a fix for that. It still was selecting outdated info (from 2023) so I added the current date into the prompt for better search results.
Week 3 full_engineering_team - I tried to expand on the coder, but it was burning through some tokens and started to cost me a pretty penny. Unfortunately, there isn't much to show for it. I think trying to get it to code a bigger front and back end was just too much for it. I might need to break down the agents even further, e.g. HTML agent, Angular agent, etc. etc.


ed-donner:
Hey! I love the weeks 1 and 2 contributions..
For the week3.. because of the way CrewAI projects work, this is adding about 15,000 lines of code to the repo! Would you be OK to remove this submission for now - just keep weeks1 and 2 in there? Thanks so much

Shmacked:
Yes, that works. Do I need to do that, or is that something you can do from your end? Kind of a Git noob, just figuring it out. It's a little unfortunate that there isn't much to show for the full engineering team. And the only real change I made to the stock picker was to include the date in the query, so I think it just makes sense to leave them out.

ed-donner:
Ahh yes you'd need to do it as far as I know.. you’d need to update the PR.. thanks so much..

_____

Edward Donner

Shmacked:
Done!

ed-donner:
Hey @Shmacked - it looks like you may have forgotten to clear outputs; there are 6,400 lines of code! Would you be OK to update the PR with cleared outputs? Many thanks

Shmacked:
Okay. It looks like I've updated it(?). In cursor, I cleared the outputs, saved it, committed it to my main branch, which means the PR should reflect that, if I understand that correctly(?).

Thanks for your patience with me!

ed-donner:
That's great! That's for sticking at it and for the contribution

PR #376, opened 2025-11-03
Added implementation for StudentCompanion Crew

Surbhit01:
StudentCompanion is a small Crewai based implementation that helps to carefully generate a study plan for the topic given by the user. There are multiple agents in the setup which are designed for specific tasks.

* `intent_recognizer` - This agent is responsible for analysing the input given by the user and returns the topic, difficulty level and additional info (if any) from the question in a structured json format.
* `topic_generator` - This agent will analyse the topic given by the previous agent and generates a comprehensive list of sub-topics to be covered for the given topic.
* `study_content_generator` - This agent will generate study content for each of the sub-topics suggested by the previous agent. While generating the content, it will also consider the difficulty level and additional info (if any) given by the user.
* `question_generator` - This agent analyses the study content generated so far and generates some questions for the user to answer, enabling a thorough understanding of the topic.
* `output_parser` - This agent will convert the response generated till now from json object to markdow for rendering.

### DIFF
--- a/3_crew/community_contributions/student_companion/.gitignore
+++ b/3_crew/community_contributions/student_companion/.gitignore
@@ -1,0 +1,3 @@
+.env
+__pycache__/
+.DS_Store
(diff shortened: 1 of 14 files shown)


ed-donner:
I love it! Nice a simple and useful. Great!

PR #381, opened 2025-11-05
Bootcamp Javi Exercise week 2

javieroteromarquez:
First I included the notebook on how to deal with images. Namely, the issue is how to deal with very long strings containing the encoded image in b64. Simply, the solution goes by storing that in a variable using a workflow approach, instead of fully agentic one.
Secondly, about the deep research, I focused on adding the questions, making an evaluator to make sure those questions were properly answered,and working with the Gradio interface so the UX worked nicely (there's streaming, there's appearing buttons and sections, etc).
[all those UX improvements are not appreciated on a screenshoot, but they are nice]
image


ed-donner:
For sure. Very nice, Javi

PR #385, opened 2025-11-06
GenAI Bootcamp - Week 4 PR Report Generator

smqd19:
This project basically builds on top of the deep research project that I had created earlier when working with OpenAI SDK. This project adds more on top of that one addresses some of the issues of that project. Below is the detailed explanation of the actual flow of the project.

The flow is as follows:

1. User inputs the topic they want a report on.
2. The clarifying node takes that topic and generates 3 clarifying questions. User then answers these questions.
3. The topic generator nodes takes this full context "the user query + the clarifying questions and their answers". Based on this, it generates 3 similar topics to broaden the scope of the research.
4. These topics generated are then fed to topic evaluator node. The topic evaluator node looks at the full context and validates the  topics generated against the full context. It generates a score between 1 - 10 on the relevance of the topic and provides feedback.
5. If the evaluator rejects the topic, the topic generator agent is called again. However, to ensure that this does not end up in an endless loop, retry count is maintained with the help of state.
6. If the maximum retry count reaches and the evaluator is still not satisfied, the topic with the best score so far is selected.
7. Now, these topics along with the full context are fed to the summarizer node. This node has the "Wikipedia Tool" attached with it. It then runs the search based on the topics and returns back the summary of the documents. The reason for using the summary was to use less number of tokens.
8. This is then given to the report writer agent. The agent takes these and generates a full report in markdown format.
9. This report is given to the evaluator alongside the full context to evaluate. If it approves, the flow moves on. Otherwise a loop similar to that of the topic evaluator follows.
10. After the report has been validated, the file writer tool writes it to the markdown file.
11. The report is then sent as email using the email tool to the user.
12. The user also receives a push notification  on their phone that email containing the report has been sent to you. This is done using the notification tool.

The code is divided in the following manner:

- The main directory has the nodes folder contains all the code files of the nodes. Creating a separate file for each node made it easy for me to make changes and keep track of the errors.
- The main directory also contains the following 4 files:
      1. The app.py files contain the Gradio UI.
      2. The tools.py file contains the tool definitions.
      3. The state.py file contains the state class.
      4. The graph_builder.py contains the graph code.
      
Below is the screenshot of the Gradio UI.
Chat

The number of files is bit on the bigger side but it helped me with keeping it all organized.

### DIFF
--- a/4_langgraph/community_contributions/muhammad_qasim_sheikh/app.py
+++ b/4_langgraph/community_contributions/muhammad_qasim_sheikh/app.py
@@ -1,0 +1,139 @@
+import gradio as gr
+import uuid
+import asyncio
+from langgraph.types import Command
+from graph_builder import research_graph
+
+# MAIN CHAT FLOW
+async def chat_flow(message: str, state: dict):
+    step = state.get("step", "START")
+    conversation_id = state.get("conversation_id")
+
+    # START THE GRAPH
+    if step == "START":
+        conversation_id = str(uuid.uuid4())
+        state["conversation_id"] = conversation_id
+        state["answers"] = []
+        state["step"] = "CLARIFYING"
+
+        yield "Got it! Let’s clarify your intent first."
+
+        config = {"configurable": {"thread_id": conversation_id}}
+        initial_input = {"user_query": message}
+
+        try:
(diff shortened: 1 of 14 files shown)


ed-donner:
Really nice @smqd19 - a great example of a full langgraph dependency graph. Great work.

PR #388, opened 2025-11-06
Add Azure OpenAI compatible version of Lab 3

YaelEfrat48:
- Fix evaluate() function to use response_format instead of response_model
- Add proper JSON schema validation for structured responses
- Include Azure OpenAI imports and SSL bypass configuration
- Maintain original functionality while ensuring API compatibility

This version works with Azure OpenAI deployments and standard OpenAI API.


ed-donner:
very useful - thanks so much for contributing this.

PR #394, opened 2025-11-08
Add Mars Lab 2 solution (cleared outputs)

marstippo:
Uploading PR for lab 2. Had chatGPT 5 mini to evaluate rankings across all models and provide it's explaination.


ed-donner:
Very nice! Outputs definitely cleared on this one.

PR #408, opened 2025-11-10
sach91 bootcamp week2 exercise

sach91:
**Week 2 exercise: using ollama models + telegram notification + evaluator model**

First generate a report from LLM based on past knowledge. Then an evaluator model checks if this report is sufficient (like for simple historical questions). Otherwise, for cases like the query needs recent information, evaluator model rejects the report and initiates web-search and then uses it to for generating report. This filters the need for invoking the time consuming web-search approach.

**Some Challenges and Insights:**
1. Free models using ollama are also competent to perform deep research.
2. It is not easy to control tool calls using Ollama for (llama 3.2) models, need to use fail-safe exceptions to control the call count from litellm level.
3. Implemented local web-search using [DuckDuckGo](https://duckduckgo.com) free search engine.

**Two examples:**
1. **Query for Historical Data: Who is the father of AI** 
    Process: Evaluator accepts the initial model report, web search not needed.
    Gradio Report:
    image
    Telegram Notification:
    image

2. **Query for Recent Information: Which is the best Machine Learning framework**
    Process: Evaluator rejects the initial model report, web search is performed.
    Gradio Report:
    image
    Telegram Notification:
    image

### DIFF
--- a/2_openai/community_contributions/sach9-deep-research-ollama/base_model.py
+++ b/2_openai/community_contributions/sach9-deep-research-ollama/base_model.py
@@ -1,0 +1,4 @@
+from agents.extensions.models.litellm_model import LitellmModel
+
+ollama_model = LitellmModel(model="ollama/llama3.2", api_key="ollama")
+ollama_model_eval = LitellmModel(model="ollama/llama3.1", api_key="ollama")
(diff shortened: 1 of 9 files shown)


ed-donner:
Very nicely done @sach91 - thank you

PR #410, opened 2025-11-10
Added Andela GenAI Bootcamp week 4 exercise - Chimwemwe

kachaje:
Created a simplified code builder that gets a request from a user to create a Python project. It uses LangGraph for state management, custom file creator and web search tools, custom local logger as well as a local Ollama instance running llama3.2 model. Attached is a slideshow of how it works. High level designs are in docs/ folder
Screen Recording 2025-11-10 at 23 21 08

### DIFF
--- a/4_langgraph/community_contributions/tochi/humandb.ipynb
+++ b/4_langgraph/community_contributions/tochi/humandb.ipynb
@@ -1,0 +1,497 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "067b892d",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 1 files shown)


ed-donner:
Hey Chimwemwe -
This is 6,500 lines of code.. I do appreciate it, but would it be possible to get this trimmed down a bit?
We probably don't need test files, READMEs and the like for a demo project, or at least not this many. Something that's tighter and easier for me to review...
Many thanks
Ed

kachaje:
I've removed the comments and docs and just updated the same PR to avoid the last experience I had with the other PR. I hope this should be better as it's the bare minimum that can demonstrate the concept. Down to 2,996 lines now

PR #413, opened 2025-11-11
Bootcamp WK4 - Sidekick Improvement [Nikhil]

rnik12:
Screenshot 2025-11-11 at 1 02 26 PM
Screenshot 2025-11-11 at 1 02 40 PM

### Features
* Added clarifier node which clarifies the original question by asking 3 sub questions.
* Using sqlite for memory so that memory is persisted in multiple runs.
* Added reattempt limits at evaluator node to prevent infinite loops.

### Key Takeaways:
* Learnt how to use Langgraph, States, Nodes and its Memory Feature
* Learnt that Evaluator Feedback Loops may run into infinite loops it is good to add limit on multiple attempts.


ed-donner:
OK very nice - thank you!

PR #414, opened 2025-11-11
Bootcamp Javi Exercise 6

javieroteromarquez:
This PR adds two main components: a sentiment analysis MCP server and a quant agent.

The MCP server uses a Hugging Face transformer (via the high-level interface) to analyze news sentiment. It now connects to an additional news API to improve coverage and relevance. The MCP handles fetching news, running sentiment analysis, and returning structured information.

The quant agent is a Python tool that downloads time series data and uses the Anthropic client as a coder to generate and execute analysis code. It runs the generated code to produce a Markdown report with results.

I'm pasting how the trace looks like when it uses these tools:
trace_executer
trace2
tracer3

Insights: The Hugging Face interface made sentiment analysis a two-lines script. Automating analysis generation with the Anthropic client was not a nice experience, it kept forgetting directories, trying weird code... overall I believe having an structured tool would have been better for this usecase.

### DIFF
--- a/6_mcp/community_contributions/exercise_week6_jom/mcp_params.py
+++ b/6_mcp/community_contributions/exercise_week6_jom/mcp_params.py
@@ -1,0 +1,52 @@
+import os
+from dotenv import load_dotenv
+from market import is_paid_polygon, is_realtime_polygon
+
+load_dotenv(override=True)
+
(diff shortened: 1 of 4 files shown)


ed-donner:
Nice experiments and fun to add your own MCP server like that.. interesting about your challenges getting claude to be reliable

PR #416, opened 2025-11-11
Bootcamp_Week_6_MCP_exercise(Cosmus_updated_version)

CosmusMutuku:
**### Exercise**

Build a Model Context Protocol (MCP)-based weather assistant that connects an LLM agent to a weather server through MCP tools.
Uses these tools — city_temp, city_condition, city_humidity, and city_wind — to fetch real-time weather data from a free weather API and present a user-friendly weather report.
The architecture includes:
- -->server.py — MCP weather server exposing tool endpoints for live data.
- -->sorrounding.py — Fetches weather data from the WeatherAPI service.
- -->weather_ui.py — Gradio-based interface that connects the LLM agent (Claude Haiku) to the MCP server via MCPServerStdio.

The result is a weather assistant capable of reasoning about conditions (e.g., “It’s raining — carry an umbrella!”).
Used OpenAI tracing and the traces are attached below

**### Challenge/Lesson**
On this revised exercise, I encountered an issue while switching from OpenAI API keys to Anthropic keys (which later became a lesson). Initially, the runner call failed with an error about invalid arguments when I tried: result = await Runner.run(agent, user_query, client). The client argument  was the problem and couldn’t process Anthropic requests properly. After some investigation, I discovered that the Anthropic integration needed two key configurations (set_default_openai_api('chat_completions'), and set_default_openai_client(client)).

Also it took me sometime/consulation to figure out how how to trace ( using set_tracing_export_api_key) in my case of having anthropic model. Learning this felt great.

**### Screenshots**
revised1

revised 2
tracer1
tracer 2
tracer 3

### DIFF
--- a/6_mcp/community_contributions/Revised_MCP/weather_ui.py
+++ b/6_mcp/community_contributions/Revised_MCP/weather_ui.py
@@ -10,17 +10,22 @@ from agents import (
     Agent,
     set_default_openai_client,
     set_default_openai_api,
+    set_tracing_export_api_key,
+    trace,
+    
 )
 
-# os.environ['OPENAI_API_KEY'] = OPENAI_API_KEY
+
 client = AsyncOpenAI(
-    #enter correct key
+    #Anthropic API key
     api_key="my_anthropic_api_key",
     base_url="https://api.anthropic.com/v1",
 )
 
 set_default_openai_api('chat_completions')
 set_default_openai_client(client)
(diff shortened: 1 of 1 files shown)


ed-donner:
OK, nicely done.
Hey I don't think you need this:
```
def query_weather(user_query: str):
    try:
        return asyncio.run(run_weather_agent(user_query))
    except Exception as e:
        return f"Error: {e}"
```
You could pass run_weather_agent directly in as a Gradio callback - it handles async fine..

PR #422, opened 2025-11-11
sach91 bootcamp week4 exercise: SideKick with ollama+docker+telegram

sach91:
**Week 4 exercise: SideKick using ollama models + telegram notification + secure python execution in docker**

Used agents and tools using LangChain / LangGraph, with the following objectives:

1. Used Docker for secure python execution.
2. Each REPL call spins up a fresh container that is destroyed after execution.
3. No persistent volume mounts; no internet access inside containers.
4. Telegram is the only outbound integration.
5. Used free models using Ollama.
6. Used LangSmith for accessing logs and calls.
7. Provide the tool response to LLMs in next interactions, not just say *Tools use*
8. Experienced that free models can handle agentic tasks to reasonable level.

**Some situations**
1. Evaluator complained about the *Tools use*
    image
2. Ask agent to calculate the square of Hubble constant (after querying information). This invokes a fresh secure docker instance and passes the code as arguments. 
    image
3. Ask to send the result via notification to telegram, got different responses based on instructions. 
    image

### DIFF
--- a/4_langgraph/community_contributions/sach91-docker-ollama-telegram/Dockerfile.pyrepl
+++ b/4_langgraph/community_contributions/sach91-docker-ollama-telegram/Dockerfile.pyrepl
@@ -1,0 +1,16 @@
+FROM python:3.12-slim
+
+# Optional: add a few safe libs you want the REPL to have
+RUN pip install --no-cache-dir numpy sympy
+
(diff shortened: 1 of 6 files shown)


ed-donner:
That's very nice @sach91 - thank you

PR #423, opened 2025-11-11
Andela GenAI Bootcamp Week4 exercise - Chimwemwe

kachaje:
Created a simplified code builder that gets a request from a user to create a Python project. It uses LangGraph for state management, custom file creator and web search tools, custom local logger as well as a local Ollama instance running llama3.2 model. Attached is a slideshow of how it works. High level designs are in docs/ folder
Screen Recording 2025-11-10 at 23 21 08


ed-donner:
Oh that's brilliant - I bet this was a lot of fun to make. Congratulations @kachaje - very nice

PR #424, opened 2025-11-11
Andela GenAI Bootcamp Week 6 Exercise - Chimwemwe

kachaje:
What's unique about my solution is that it uses a local web search tool and local tracing. It also uses a local LLM using Ollama and llama3.2. Attached are sample screenshots
Screenshot 2025-11-11 at 18 51 34
Screenshot 2025-11-11 at 18 52 08


ed-donner:
OK, that's great! I hope it makes you lots of virtual money!

PR #434, opened 2025-11-12
Solisoma week6

solisoma:
# BootCamp PR Week6: Sentiment Trader

## What I Built

Extended Day 5's Autonomous Traders by adding:
- **Custom MCP Server** (`sentiment_server.py`) - 3 tools for real-time news sentiment analysis
- **5th Trader: Solisoma** - Sentiment-driven momentum trader that checks news before every trade

### 1. **MCP Servers Are Surprisingly Simple**
I thought building a custom MCP server would be complex, but FastMCP made it straightforward. The hardest part wasn't the MCP protocol it was making the **web scraping robust**.

### 2. **DRY vs. Pragmatism**
Initially tried to reuse parent files to avoid duplication. This created massive complexity - import order issues, path problems, and fragile dependencies. Spent hours debugging "Connection closed" errors just from wrong working directories.
Final decision: Copied all needed files into solisoma/ folder. Yes, it duplicates code. But it's self-contained, won't break when parent files change, and actually works reliably.

### 3.  **API Mistake**
Found a bug in `market.py` - the `is_market_open()` function only checked for `"open"`:
```python
return market_status.market == "open" 
```
But Polygon API returns **three states**: `"open"`, `"extended-hours"`, and `"closed"`. The original code prevented trading during pre-market (4-9:30 AM) and after-hours (4-8 PM) even though these are valid trading windows.
**Fix:**
```python
return market_status.market in ["open", "extended-hours"]
```

### DIFF
--- a/6_mcp/community_contributions/solisoma/sentiment_server.py
+++ b/6_mcp/community_contributions/solisoma/sentiment_server.py
@@ -208,3 +208,6 @@ async def analyze_overall_sentiment() -> str:
     except Exception as e:
         return f"Error analyzing market sentiment: {str(e)}"
 
+
+if __name__ == "__main__":
+    mcp.run(transport='stdio')


ed-donner:
Hmm - I don't understand how the sentiment mcp server could be working. You don't appear to ever call mcp.run() in that module, so it can't be running?

solisoma:
I have updated it 
I somehow ommitted it

ed-donner:
OK sounds good - thanks much

PR #437, opened 2025-11-12
Kwabena bootcamp week 2

kbaah:
deep research

i think openai's agent sdk is amazing! try out my deep research tool which gathers clarifications before going off to research


ed-donner:
Very nice - thanks Kwabena

PR #439, opened 2025-11-12
Kwabena Bootcamp Week4

kbaah:
sidekick

Incorporating user memory and session management proved time consuming but that would be a good thing to add in future.


ed-donner:
OK many thanks

PR #447, opened 2025-11-12
(Oct 2025 Bootcamp): Add AutoTrader components including accounts management, trading logi…

hopeogbons:
# AutoTrader

trades

### Original Problem Identified:
I observed that in the legacy implementation, each trader spawned its own set of MCP server subprocesses independently. Specifically, each of the 4 traders created 8 separate subprocesses per trading cycle:

- 2 temporary subprocesses when the application code called accounts_client functions to read account data and strategy
- 6 persistent subprocesses for the trader agent's MCP servers (3 trader servers: accounts, push, market; and 3 researcher servers: fetch, brave search, memory)

This resulted in 32 total subprocesses per cycle (8 × 4 traders), with significant duplication since all traders needed access to the same accounts, push, and market servers.

### Solution Implemented:
I refactored the application to use an orchestrator pattern with shared MCP server connections. The OrchestratorAgent class (which is a resource management coordinator, not an AI agent) now:

- Creates shared trader MCP servers once (3 servers: accounts, push, market) that all 4 trader agents access simultaneously
- Creates dedicated researcher MCP servers per trader (2-3 servers each) since they require separate memory databases for knowledge isolation
- Passes these pre-initialized connections to each trader agent instead of having them spawn their own

### Performance Improvement:
- Before: 32 subprocesses (8 per trader × 4 traders)
- After: 11-15 subprocesses (3 shared trader servers + 2-3 researcher servers per trader)
- Reduction: ~75% fewer subprocess spawns

This architectural change significantly reduces process forking overhead, improves resource utilization, and accelerates overall system performance by eliminating redundant MCP server instances while maintaining full functionality and agent autonomy.

### DIFF
--- a/6_mcp/community_contributions/AutoTrader_hopeogbons/accounts.py
+++ b/6_mcp/community_contributions/AutoTrader_hopeogbons/accounts.py
@@ -1,0 +1,186 @@
+from pydantic import BaseModel
+import json
+from dotenv import load_dotenv
+from datetime import datetime
+from market import get_share_price
+from database import write_account, read_account, write_log
+
+load_dotenv(override=True)
+
+INITIAL_BALANCE = 10_000.0
+SPREAD = 0.002
+
+
+class Transaction(BaseModel):
+    symbol: str
+    quantity: int
+    price: float
+    timestamp: str
+    rationale: str
+
+    def total(self) -> float:
+        return self.quantity * self.price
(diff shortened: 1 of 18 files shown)


ed-donner:
Very interesting point, Hope! Thank you for that - nicely done - and an interesting approach

PR #448, opened 2025-11-12
GenAI Bootcamp - Week 6 PR Content Quality Auditor

smqd19:
The **Content Quality Auditor** is an MCP-compliant agent that evaluates a piece of content for readability, SEO effectiveness, tone, and plagiarism risk.
 It is built on top of the OpenAI Agents SDK and FastMCP, showing a complete end-to-end flow where:

1. The MCP server exposes modular tools.
2. The LLM agent connects to that server dynamically.
3. A lightweight UI lets the user submit a URL or HTML text for auditing.
4. The analysis results are returned as both a human-readable report and a structured JSON record saved to a local database.

**File-Level Explanation**
1. `seo_audit_server.py`
 Defines the MCP server and all modular tools (`compute_readability`, `compute_seo`, `compute_tone`, `compute_plagiarism`, and `full_audit`).
 Each tool runs independently and returns structured data. The server operates via STDIO so it can be discovered and used dynamically by the agent without any hardcoded imports.
6. `auditor.py`
 Implements the LLM agent that connects to the MCP server at runtime.
 It uses the `Agent`, `Runner`, and `MCPServerStdio` classes from the Agents SDK to send the user’s query to the model, allow it to call the MCP tools, parse the JSON output, and store the results in the database.
 This file is the orchestration layer where the model’s reasoning and the server’s tool execution meet.
7. `database.py`
 Creates and manages a simple SQLite database (`audits.db`).
 Every audit result, along with its metadata and JSON report, is written automatically for recordkeeping and later review.
8. `audit.py`
 Defines the structured `AuditResult` dataclass used for consistent report formatting and scoring.
 It ensures all outputs share the same schema and grading logic.
9. `app.py`
 Provides a clean Gradio-based front end for interacting with the auditor.
 Users can enter a URL or paste HTML, trigger the audit, and immediately see the generated report.
 It calls the `run_audit_agent()` function from `auditor.py` and displays the model’s Markdown summary and structured findings.

image
image (1)

### DIFF
--- a/6_mcp/community_contributions/muhammad_qasim_sheikh/app.py
+++ b/6_mcp/community_contributions/muhammad_qasim_sheikh/app.py
@@ -1,0 +1,55 @@
+import gradio as gr
+import requests
+from auditor import run_audit_agent
+
+
+DEF_HTML = """<html><head><title>Sample</title><meta name="description" content="Demo article"></head>
+<body><h1>Hello world</h1><p>This is a simple demo paragraph that is fairly easy to read. It avoids complex structures.</p></body></html>"""
+
+def fetch_url(url: str) -> str:
+    if not url:
+        return ""
+    try:
+        r = requests.get(url, timeout=15)
+        r.raise_for_status()
+        return r.text
+    except Exception as e:
+        return f"Error fetching URL: {e}"
+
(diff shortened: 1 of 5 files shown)


ed-donner:
Yep, that works. Thank you.

PR #449, opened 2025-11-12
sach91 bootcamp week6 exercise: MCP with ollama+docker+telegram

sach91:
## MCP using ollama models + telegram notification + docker container

First generate a report from LLM based on past knowledge. Then an evaluator model checks if this report is sufficient (like for simple historical questions). Otherwise, for cases like the query needs recent information, evaluator model rejects the report and initiates web-search and then uses it to for generating report. This filters the need for invoking the time consuming web-search approach. The process uses calls to MCP server for managing various operations.

```mermaid
flowchart TD
    A[Initial Report]
    B[Eval Model]
    C[Plan Search]
    D[Perform Search]
    E[Summarize]
    F[Timestamp]
    G[Send Message]

    A --> B
    B -->|Fail| C
    C --> D --> E --> F --> G
    B -->|Pass| E

    %% Make Eval Model red
    style D fill:#ffcccc,stroke:#ff0000,stroke-width:2px
    style F fill:#ffcccc,stroke:#ff0000,stroke-width:2px
    style G fill:#ffcccc,stroke:#ff0000,stroke-width:2px
```
*Red Blocks support MCP server calls for providing the functionality.* 

### Some Challenges and Insights
1. Free models using ollama are also competent to perform deep research.
2. It is not easy to control tool calls using Ollama for (llama 3.2) models, need to use fail-safe exceptions to control the call count from litellm level.
3. Implemented local web-search using [DuckDuckGo](https://duckduckgo.com) free search engine.
4. Used Docker for secure python execution to get timestamp information, which is added in the response.
5. Used OpenAI Traces for accessing logs and calls.
6. Telegram Notification, Web Search and Docker call are managed through MCP server.
7. Experienced that free models can handle agentic tasks to reasonable level.
8. Amazed to see Google Gemini goofing up (see below). 

### Some situations

1. **Query for Historical Data: Who was Alan Turing** 
    Process: Evaluator accepts the initial model report, web search not needed.
    Gradio Report:
    image
    Telegram Notification:
    image

2. **Query for Recent Information: Who won Turing Award in 2025**
    Process: Evaluator rejects the initial model report, web search is performed.
    Gradio Report:
    image
    Telegram Notification:
    image

3. It should be noted that the model knowledge cut-off date was December 2023, and the above information was only feasible due to the web-search.
    image

4. Strangely, asking the same question to Google goofed Gemini, which marked the 2024 winners as of 2025. 
    image

    I confirmed from ACM website that our answer is correct:
    https://www.acm.org/media-center/2025/march/turing-award-2024

### DIFF
--- a/6_mcp/community_contributions/sach91-mcp-ollama-telegram/Dockerfile.pyrepl
+++ b/6_mcp/community_contributions/sach91-mcp-ollama-telegram/Dockerfile.pyrepl
@@ -1,0 +1,12 @@
+FROM python:3.12-slim
+
+# Optional: add a few safe libs you want the REPL to have
+RUN pip install --no-cache-dir numpy sympy
+
+# Create an unprivileged user
+RUN useradd -m -u 1000 sandbox
+USER sandbox
+WORKDIR /home/sandbox
+
+# Keep Python quiet and unbuffered
+ENV PYTHONDONTWRITEBYTECODE=1 PYTHONUNBUFFERED=1
--- a/6_mcp/community_contributions/sach91-mcp-ollama-telegram/app.py
+++ b/6_mcp/community_contributions/sach91-mcp-ollama-telegram/app.py
@@ -1,0 +1,23 @@
+import gradio as gr
+from dotenv import load_dotenv
+from research_manager import ResearchManager
+
+load_dotenv(override=True)
+
+
+async def run(query: str):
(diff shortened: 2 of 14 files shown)


ed-donner:
Gotcha - simple and nicely integrated with a prior project. Thanks.

PR #457, opened 2025-11-13
Add OpenAI projects: MedReach and AskMyCV

lisekarimi:


ed-donner:
Thanks so much @lisekarimi !

PR #461, opened 2025-11-14
Two new practice projects. One exploring tolls and prompts and the other exploring tools, guardrails and handoffs.

glafrance:


ed-donner:
Thanks so much @glafrance - a ton of work here

PR #474, opened 2025-11-23
Feature: Added example for Week 1 Day 3 (Medical Note Agentic Workflow with Evaluation)

narquette:
# Overview

- Added a new community example for week 1 day three using note text from mTSamples.
- The agentic workflow asks different models to look at the note and output mentioned diagnoses with the corresponding ICD10CM note information with another model performing a ranking each model output.

# Other Notes

- Pydantic-settings can be something that is used in other notebooks to store the API keys. The class will automatically look for mentioned environment values and override the default values.
- Enumerate can also have a start index (start=1)


ed-donner:
Very nice! Thank you

PR #479, opened 2025-11-26
contribution: week2: drawing agents

sirknigget:
Project that utilizes OpenAI Agents SDK to plan a drawing on a canvas, execute the drawing using an exposed tool and then evaluate it using vision input to a model.


ed-donner:
That's a really cool project! Thank you for sharing

PR #483, opened 2025-11-30
Add basics of MS Agent Framework (ex-autogen) and Sidekick built on it

lisekarimi:


ed-donner:
Brilliant! very helpful. Thanks so much @lisekarimi

PR #484, opened 2025-12-01
Next business day tool addition lab 2 week 2(outputs cleared)

AnnpaS18:
Added a tool that will get the next business day, which is added in the email body by the email manager agent.


ed-donner:
Nice - thanks!

PR #487, opened 2025-12-03
Add asyncio research notes

threecuptea:
To summarize
- 'await asyncio.sleep(secs)' block (pause) the current coroutine call and let event loop to run other coroutines while it is waiting. However, it is not blocking the entire thread like 'time.sleep()' does.  If time.sleep() in the main thread, time.sleep() will halt the whole process flow. That is not desirable.
- If you are doing
> await do_some_work('Sonya')
    await do_some_work('Jimmy', 2)
    await do_some_work('Peter', 3)
>    
    
You would see the same result like
>sync_work_with_time_sleep('Sonya')
sync_work_with_time_sleep('Jimmy', 2)
sync_work_with_time_sleep('Peter', 3)
>

with the output of 'Starting work, Work complete' pairs because there is only one task  in the event loop and not much to yield to.

`asyncio.gather` would give more tasks in event loop and allow parallelism.  The output would be like 
> Overall start time: 1764789322.65409
Starting work for Sonya
Starting work for Jimmy
Starting work for Peter
Work complete after 1 second(s), at 1764789323.656093
Work complete after 2 second(s), at 1764789324.656141
Work complete after 3 second(s), at 1764789325.656134
Time taken: 3.0026931762695312 seconds
>

vs using sync time.sleep()  
>Overall start time: 1764789604.758095
Starting work for Sonya
Sync Work complete after 1 second(s), at 1764789605.763951
Starting work for Jimmy
Sync Work complete after 2 second(s), at 1764789607.768606
Starting work for Peter
Sync Work complete after 3 second(s), at 1764789610.773984
Time taken: 6.016453981399536 seconds
>

Also other than `asyncio.run()` to start even loop, we can use context to start even loop as the followings:
> with asyncio.Runner() as runner:    
        runner.run(do_a_lot_of_work_in_parallel())
>

### DIFF
--- a/guides/community_contributions/sonyaling1947_asyncio_notes.ipynb
+++ b/guides/community_contributions/sonyaling1947_asyncio_notes.ipynb
@@ -1,0 +1,230 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "5c291475-8c7c-461c-9b12-545a887b2432",
+   "metadata": {},
+   "source": [
+    "# Async Python\n",
+    "\n",
+    "## A briefing on asynchronous python coding, essential in Agent engineering"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "538fa044",
+   "metadata": {},
+   "source": [
+    "Here is a masterful tutorial by you-know-who with exercises and comparisons.\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


ed-donner:
Thank you!

PR #494, opened 2025-12-09
Add skruglewicz lab 2 Python solution

skruglewicz:
# Multi-Model Evaluator (2_lab2.py)

A Python script that evaluates and compares the performance of multiple AI language models by generating a challenging question, collecting responses from various providers, and ranking them using a judge model.


ed-donner:
Great! Thank you

PR #502, opened 2025-12-16
POC: Software team crewai flow with UI with Agents not testing it with docker.

wieczoreksdev:
Hello Ed,

Time to contribute to your “tiny” 😉 repo!
The app allows users to run it directly from UI (ui.py), provide requirements and modules, and have CrewAI Flow generate a packaged application automatically. The final result? Users can download the app from the UI.

I used CrewAI Flows to handle multiple modules - easiest and safest approach. I experimented with hierarchical processes, but a heads-up: the Manager agent doesn’t save tokens, so credits disappear fast. ☕. I think to test hierarchical process it is the best to use a local Ollama model with CrewAI .

Also tested the router - works well when strings are returned from it, but when introduced e2e engineer almost at the end of the flow so that it can test if the code is reliable, it was not possible to come back the beginning of the flow by returning task name.

I also encountered an issue with the zipping. It was done by the Agent but on Windows machine it was not possible to open zipped file. So my project can be treated purely as poc that can be expanded.

App specifics:
Type: CrewAI Flow
Tools: Custom Zip Tool, SerperDevTool, FileReadTool
Additional Agents: Business Analyst (.puml diagrams), Zip Engineer (zips the app)
Flow Additional Features: Uses router (watch out for task-name return quirks)
Models Used: openai/gpt-4o-mini, llama3.2, anthropic/claude-haiku-4-5-20251001
No docker code testing applied --> Agents write and do not run it - sweet ride.

Hope you enjoy this tiny summary as much as I enjoyed building the app and it! 😄.

And Merry Christmas

Best Regards,
Seb


ed-donner:
Hey.. hmm. well this would be quite the christmas present, as it appears to add 42,000 lines of code across 215 files. Is that intentional?

wieczoreksdev:
omg.
thank you for spotting that.
I will close and open a new one.

PR #504, opened 2025-12-18
community contribution: slogan generator (an evaluator-optimizer pattern example)

ReubenBeeler:
publishing my 2_lab2 exercise for the community! I quite like my slogans for wet socks...


ed-donner:
nice!

PR #505, opened 2025-12-18
Add SendGrid inbound reply agent with agentic SDR flow

urvashicodess:
This contribution adds a working example of an agentic SDR reply flow using SendGrid inbound webhooks.

Features:
- Phase 1 outbound email sent via agent selection
- Dynamic reply-to addresses per lead
- FastAPI webhook to receive SendGrid inbound replies
- SDR reply agent that responds contextually and sends a follow-up email
- Clean separation between side effects (email) and agent reasoning

This example demonstrates a realistic end-to-end agent workflow:
outbound → inbound → agent reasoning → reply.

Happy to adjust structure or naming if needed.


ed-donner:
That's fantastic - many thanks

PR #506, opened 2025-12-20
Deep Research: using EmailJS, Serper Google search, Azure Foundry LLMs

f2002642:
Replace SendGrid with EmailJS, switch to Serper Google search, use Azure Foundry LLMs, and disable openai tracing..


ed-donner:
Great - thank you so much

PR #513, opened 2025-12-22
Mutual fund selector and report generator- Deep research(cleared o/p)

AnnpaS18:
This is a mutual fund selector and report generator agent- It displays top 5 mutual funds and chooses the best out of it and generates a report on it based on the query and its respective clarifying questions.
The repost can also be sent as an email


ed-donner:
Thanks for the contribution!

PR #514, opened 2025-12-22
Add sidekick Juan's contribution

Jsrodrigue:
Hi Ed,

This is my Sidekick project built with LangGraph. I included a simple login system and a RAG-based workflow with the possibility of adding and selecting multiple folders as independent knowledge bases. Each folder maintains its own isolated context and persistent conversation.

I hope you can take a look at it and share any feedback. 

Thank you!


ed-donner:
That's great - nice features for Sidekick - thanks

PR #520, opened 2025-12-24
langgraph sidekick, a working HuggingFace docker deployment with Playwright chromium 

threecuptea:
It's quite a journey to get Sidekick docker deployment with Playwright (Chromium) and Gradio working.  Share my experience.


ed-donner:
Goodness yes - that is not an easy task at all. Congrats on making it happen! Thank you for sharing.

PR #521, opened 2025-12-25
updated my segments of foundations/community_contributions

jolugbo:
I included a folder with my name (jolugbo_bots) to hold my agent, the configuration generates a question from one model, and passes it onto another model to resolve.


ed-donner:
Nicely done, and thank you for keeping everything tidy.. thanks for sharing!

PR #527, opened 2025-12-27
Shawn lab2 contribution 1

Shawnarya25:
Made some necessary changes. When you are using open router API keys, how can you work efficiently


ed-donner:
Ah and this one too please

PR #533, opened 2026-01-03
Fix polygon not authorized error, 'market' table does not exist error and Max turns (10) exceeded on `Researcher` error even though MAX_TURNS = 30 is configured

threecuptea:


ed-donner:
Wow this is brilliant @threecuptea 
Looks like you've finally solved the timezone polygon problem for reals. If you wanted to make another PR that makes this isolated change in the main codebase (i.e. outside the community_contributions) then I'll gratefully merge it, or I'll do it myself sometime. Thank you for this excellent work!
Ed

threecuptea:
@ed-donner Do you want me include all 3 changes or just `Polygon not authorized error fix` into the main codebase?
1) Polygon not authorized error fix 
2) market table not exists (switch to use class Database)
3) MAX_TURN applied to research_tool.

PR #538, opened 2026-01-07
Create 3_lab3_fix.ipynb

sankargh:
If we face a TypeError while running 'gr.ChatInterface(chat, type="messages").launch()' method, we can use below method as alternate 
gr.ChatInterface(fn=chat).launch()


sankargh:
Updated weeek 3 lab 3 with a simple fix

PR #549, opened 2026-01-15
MCP: add aiobot with MCP Weather, Events API

lisekarimi:


ed-donner:
Ahh thank you @lisekarimi - very clear and nice.

PR #559, opened 2026-01-19
Windows (no-wsl) demo for Jupyter and Python

miguelremi:
This PR enables MCP (Model Context Protocol) servers to work natively on Windows without requiring WSL, specifically addressing tool compatibility issues in Jupyter notebooks.

This solution could be particularly useful for users who want to work through the labs without installing the full WSL environment on Windows.

The demo supports both OpenAI cloud models and local Ollama instances, with custom tracing for local execution.

### DIFF
--- a/6_mcp/community_contributions/windows_no_wsl/.gitignore
+++ b/6_mcp/community_contributions/windows_no_wsl/.gitignore
@@ -1,0 +1,2 @@
+# Sandbox directory for MCP filesystem server
+sandbox/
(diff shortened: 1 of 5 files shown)


ed-donner:
Wow thank you.. does this work reliably? Everyone seems to give different answers on stdio on Windows, even with patches like this?! I'm super happy to hear if this works for you reliably. Thank you for sharing

miguelremi:
Thanks so much, Ed! Yes, it's been working and usable for me. I tested it with Python 3.12.12 across labs 1, 2, 3, and 4 in 6_mcp (lab 5 is pure Python so should work without patching with recent Python versions).

Memory tools worked well with gpt-4o (though gpt-4o-mini needed a more direct prompt to use tools in lab 3). I also tested the Polygon API integration just to verify the tool calls were going through correctly (got the expected "upgrade your plan" response 😄).

Just note this is specifically for playing with the notebooks from Windows - it won't work with Node.js MCP servers that don't use the npx command to run.

Thank you for the LLM Engineering and Agents courses. Really engaging stuff!

PR #560, opened 2026-01-19
3_lab3 using Claude evaluator

jeremyw76:
- The Anthropic API for structured outputs is still in beta and requires version 0.76.0 of the Anthropic SDK.
- I'm not a Python developer, so it's possible that there's a better way to use the Pydantic model with the Anthropic SDK. This works, though.
- Claude rejects messages that contain pig Latin. This is because it considers pig Latin to be a security risk: pig Latin may be used in an attempt to bypass Claude's guardrails.
- I experimented with instructing Claude to accept pig Latin, given that this is a training exercise. I was not able to convince it.
- I also thought about handling the Claude response so that a refusal simply resulted in a failed evaluation. I abandoned this idea because it really defeats the purpose of the exercise.
- I resorted to changing the system message for the first LLM to act like a baseball coach rather than using pig latin. It's a simple solution, and it's also hilarious.

### DIFF
--- a/1_foundations/community_contributions/3_lab3_claude_evaluator.ipynb
+++ b/1_foundations/community_contributions/3_lab3_claude_evaluator.ipynb
@@ -1,0 +1,388 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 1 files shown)


ed-donner:
Hahaha brilliant! Thank you so much @jeremyw76

PR #568, opened 2026-01-23
1_Foundations Lab5 exercise

RajuDasa:
Agent loop exercise - stock recommendation based on past performance.


ed-donner:
Ooh the first Agent Loop exercise! Thank you!

PR #570, opened 2026-01-25
Did Lab 1 and Lab2 with gemini API.

thiagoporfirio86:


ed-donner:
Hey - I'd prefer to use the approach documented in Guide 9 for this, as covered in the course and mentioned at the top of Day 1 Lab 1 by the big stop sign.. that will be consistent and compatible with all projects..

PR #574, opened 2026-01-28
rewrote career chatbot to use OpenAI

vaskod924-lang:
This is a Jupyter notebook containing the career chatbot code from 4_lab4.ipynb rewritten by using OpenAI SDK.


ed-donner:
Hmm isn't the lab4 notebook already using OpenAI? What did you change here again?
Also, it looks like Outputs might be saved?

PR #583, opened 2026-02-07
feat: Update Polygon to Massive due to MCP rename

SatLibRe:
This PR updates all references from Polygon.io to Massive following their recent rebrand in `3_lab3` of the MCP section.

While a rebrand alone would not normally require a full update, Massive has also renamed and updated their MCP repository, along with related server configuration and environment variables. Specifically, the MCP server is now sourced from:

`uvx --from git+https://github.com/massive-com/mcp_massive@v0.6.0 mcp_massive`

When I tried to use the old setup, it failed. This changes fixes that.

This PR addresses those changes by:

- Updating the MCP server to the latest stable version from Massive

- Revising references to Polygon.io in `3_lab3` of the MCP section of the course to reflect the new naming

To avoid introducing breaking changes, `polygon-api-client>=1.14.5` has intentionally been left in the .toml file.

Note: This change will require either a re-recording or a short supplemental video to explain the updates and keep the course content aligned with the code.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,6 +20,7 @@ dependencies = [
     "langgraph-checkpoint-sqlite>=2.0.6",
     "langsmith>=0.3.18",
     "lxml>=5.3.1",
+    "massive>=2.2.0",
     "mcp-server-fetch>=2025.1.17",
     "mcp[cli]>=1.5.0",
     "openai>=1.68.2",


ed-donner:
I've just run the code precisely as it is in the repo, with the original Polygon MCP server name, and the polygon key, and it ran fine and returned Apple's stock price.
When you say, "When I tried to use the old setup, it failed." - did you try to debug the issue? Did you identify the root cause of the issue before trying the Massive server?
Many thanks
Ed

PR #586, opened 2026-02-11
Add langgraph-sidekick contribution

TiagoIesbick:
My SideKick version. Hope you like it.


ed-donner:
This is great! Thank you! Would you mind pruning this down a bit to make it best for a community_contribution - 4K lines of code is a lot to review..
- Make READMEs concise or remove
- Remove database files and other binaries that can be generated
- Remove uv.lock if it can be generated
- Remove a LICENSE file as the repo is already MIT
Some of the prompts seem on the verbose side. Was some of this suggested by LLMs? it might be worth doing a round of asking them to make it more concise - you often get the same or better performance.
Hope that's OK.. I love adding Sidekicks to the repo, but slim PRs are always appreciated..
Thanks so much
Ed

TiagoIesbick:
Hi Ed,

Thank you for the thoughtful feedback — I really appreciate it.
I’ve made the following updates:
- Pruned the repository to reduce overall size
- Simplified the README to make it more concise
- Removed generated artifacts (database files and other binaries)
- Removed uv.lock since it can be generated
- Removed the LICENSE file (as the repository already uses MIT)
Regarding the prompts: I did experiment with making them more concise (including multiple refinement rounds), but the current versions were the ones that consistently produced the most reliable behavior in this setup. For that reason, I kept them as-is for now.
Thanks again for the review and for maintaining the Sidekicks initiative — I appreciate the guidance.

Best,
Tiago

ed-donner:
Makes total sense - thanks so much

PR #587, opened 2026-02-11
Feature | Add sidekick contribution (Planning and Engagement)

GastonAQS:
This PR adds a new community contribution: sidekick_with_planning.

It extends Sidekick with a planning-first workflow, plan evaluation before execution (also the plan is validated), and a post-success engagement agent that asks brief follow-up questions only when useful. If no follow-up is needed, the agent returns NO_QUESTIONS and nothing is appended.


ed-donner:
Thank you!

PR #589, opened 2026-02-11
Created an agentic employee onboarding tool

Shield01:


ed-donner:
A great use case - thanks!

PR #591, opened 2026-02-12
Padmanabh_2_lab_2_fallback_flexible_providers_option

padmanabh275:
Added a fall back option for users who do not have any LLMs working. They can try every Key out and whichever is working the program uses it to compare.


padmanabh275:
Added a fall back option for users who do not have any LLMs working. They can try every Key out and whichever is working the program uses it to compare

PR #602, opened 2026-02-15
hotfix: Fix agent-not-found errors: only route refinement to registered agents

gsarchioni:
## Problem
When an agent tried to bounce an idea off another agent, the runtime could log:
`ERROR:autogen_core:Agent agent2 not found, failed to deliver message.`

`find_recipient()` was choosing from every `agent*.py` file on disk. With many concurrent requests, some agents (e.g. agent20) are registered and run before others (e.g. agent2). Picking agent2 before it was registered caused the delivery failure.

## Solution
- **Track registered agents**: Introduced `REGISTERED_AGENTS` and `register_agent()` in `messages.py`. The Creator calls `register_agent(agent_name)` after each successful registration with the runtime.
- **Route only to registered agents**: `find_recipient()` now selects only from `REGISTERED_AGENTS` instead of from all `agent*.py` files.
- **Exclude self and handle no recipient**: `find_recipient(exclude=...)` excludes the current agent so we don’t send to ourselves. It returns `None` when there is no valid recipient; the agent then skips the bounce and returns its idea as-is.
- **Cleanup**: Removed the unused (and broken) `from agent import Agent` in `world.py`; the module only defines `AgentCreator`.

## Files changed
- `messages.py` – `REGISTERED_AGENTS`, `register_agent()`, `find_recipient(exclude)` with registered-only selection
- `creator.py` – call `messages.register_agent(agent_name)` after registering an agent
- `agent.py` – use `find_recipient(exclude=self.id.type)` and only bounce when `recipient is not None`
- `world.py` – remove invalid `Agent` import

### DIFF
--- a/5_autogen/agent.py
+++ b/5_autogen/agent.py
@@ -8,13 +8,12 @@ from dotenv import load_dotenv
 
 load_dotenv(override=True)
 
-class Agent(RoutedAgent):
 
-    # Change this system message to reflect the unique characteristics of this agent
+class AgentCreator(RoutedAgent):
 
     system_message = """
     You are a creative entrepreneur. Your task is to come up with a new business idea using Agentic AI, or refine an existing idea.
-    Your personal interests are in these sectors: Healthcare, Education.
+    Your personal interests are in these sectors: Consulting, Insurance.
     You are drawn to ideas that involve disruption.
     You are less interested in ideas that are purely automation.
     You are optimistic, adventurous and have risk appetite. You are imaginative - sometimes too much so.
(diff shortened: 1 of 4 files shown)


gsarchioni:
Hi @ed-donner - please review this PR. I noticed an issue with the `autogen` final project when an `agent*.py` created but have not been registered to the runtime before another agent calls it.

ed-donner:
Oh gosh - this seems like a great fix, but would you mind putting it in community_contributions? At some point in the future I will test it and look to bring it into the main repo. Many thanks

PR #612, opened 2026-02-19
Update model version from gemini-2.0-flash to gemini-2.5-flash

pooyas:
gemini-2.0-flash is no longer available. updating it to point to gemini-2.5-flash


ed-donner:
Ah - many thanks

PR #626, opened 2026-02-27
Add Amar's lab 2 solution(cleared outputs).

amargurung1013:


ed-donner:
Great - thanks so much!

PR #673, opened 2026-03-17
Modified version of 'debate' crew with 5 agents

zoya-hammad:


ed-donner:
hey.. would it be possible to remove uv.lock from this, and see if there's any way some files could be reduced to make the PR more concise? Thanks so much

PR #970, opened 2026-03-25
Week 6 exercise

CynthiaOmovoiye:
````markdown
# Curiosity Break MCP

A tiny MCP server with 3 simple tools that do not need any API key.

## Tools

- `random_advice` → gives one random advice
- `xkcd_latest` → gets the latest XKCD comic
- `xkcd_by_number` → gets a specific XKCD comic by number



---

## Quick option: use the live URL

This MCP is already live here:

```text
https://curiosity-break-mcp.onrender.com/mcp
````

So anyone can connect to it from Cursor without running the server locally.

Example Cursor config:

```json
{
  "mcpServers": {
    "curiosity-break-live": {
      "url": "https://curiosity-break-mcp.onrender.com/mcp"
    }
  }
}
```

## Credits

* Advice Slip API
* XKCD JSON API

### DIFF
--- a/6_mcp/community_contributions/curiosity_break_mcp/README.md
+++ b/6_mcp/community_contributions/curiosity_break_mcp/README.md
@@ -1,0 +1,205 @@
+
+````markdown
+# Curiosity Break MCP
+
+A tiny MCP server with 3 simple tools that do not need any API key.
+
+## Tools
+
+- `random_advice` → gives one random advice
+- `xkcd_latest` → gets the latest XKCD comic
+- `xkcd_by_number` → gets a specific XKCD comic by number
+
+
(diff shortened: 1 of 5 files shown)


CynthiaOmovoiye:
Hi @hopeogbons Kindly review. Thank you

hopeogbons:
Great job! #LGTM 👍 

Note: 
- The dashboard makes two live external API calls on every page load with no fallback. If either `adviceslip` or `xkcd` is down, the whole page throws a 500 instead of degrading gracefully.
- `result.content[0].text` assumes the first content block is always text. No type check, so it'll blow up if the server ever returns a non-text content block.

PR #1285, opened 2026-04-09
fix - w3_engineering_team change to claude-sonnet-4-5 from claude-3-7-sonnet which is obsoleted

mihaly-hazag:
(claude-sonnet-4-6 does not work with the current crewai version)


ed-donner:
thanks so much for this!

PR #1286, opened 2026-04-09
Stock Picker project with LMStudio

AdityaNarendra:
The stock picker project uses LMStudio for LLM access and embeddings. Modified the lmstudio_loader.py to present a crewai focused embedding configuration dictionary based on OpenAI as the provider.


ed-donner:
Great - thanks so much!

PR #1288, opened 2026-04-10
Community: LangGraph three-agent debate with dynamic routing

srinivasrk:
### Title
**Community: LangGraph three-agent debate (dynamic routing + debiasing)**

### Description

Adds `4_langgraph/community_contributions/three_agent_debate_langgraph/three_agent_debate.ipynb`: a **for / against / moderator** debate on a shared topic, orchestrated as a LangGraph workflow.

**Pattern:** Supervisor-style **moderator** routes turns; two debater nodes produce arguments. **Runtime routing** via `Command(goto=..., update=...)` (not only static edges). Shared **`DebateState`** with **list reducers** (`Annotated[list[str], add]`) for growing transcripts. **`MemorySaver`** checkpointer. Moderator/judge use **structured outputs** (Pydantic) for decisions and final verdict. **Optional Step 6** (`app.stream(..., stream_mode="updates")`) is included but **commented out** in the notebook—uncomment that cell if you want live node-by-node output.

**Anchoring bias:** LLMs can overweight early arguments or a one-sided view. The graph **always opens with `for_agent` → `against_agent`** so the moderator’s first pass happens only after **both** openings exist; it **requires at least one follow-up** before allowing an early end, and caps moderator rounds with a **forced winner/tie** when stopping.

**Dependencies:** Adds `langchain-google-genai` to `pyproject.toml` so `init_chat_model(..., model_provider="google_genai")` works (run `uv lock` / `uv sync` locally to refresh `uv.lock` if needed).

**Run:** Set `GEMINI_API_KEY` (and optionally `GEMINI_MODEL`) in `.env`.


ed-donner:
Hi - thank you for the contribution. It's not practical or feasible to change dependencies for the entire repo. With 200,000 people on the course, and the need for all labs to work on all systems, merging this PR would involve very significant retesting. If you have additional project dependencies, simply note that in your contribution; others can run "uv add" themselves. Thanks so much

srinivasrk:
Makes sense I will remove the dependency and resubmit the PR. Thanks!

ed-donner:
Great - thanks so much!

PR #1298, opened 2026-04-12
Add Rdkprop Lab 4 travel planner app (1_foundations)

Rdkprop:
Created a Travel planner app on Huggingface


ed-donner:
Great - thanks so much!

PR #1299, opened 2026-04-12
Added working example to use mcp with rag

himanshuramranjan:
Have added a examples on how to use RAG along with mcp and agents (using groq model). I have added one example to use without storing the data (by keeping it in code itslef), and in second example I have used ChromaDB to store the data locally.


himanshuramranjan:
Hi @ed-donner, can you please review and merge this code. From my understanding its a basic but good example to integrate RAG with MCP learning

I have also added the steps to do any necessary installations and setup

ed-donner:
Great - thanks so much!

PR #1304, opened 2026-04-16
Add resume revision multi-model script

soflatechie:
This script is using the idea of 2-lab-2, where we
 ask the same question to multiple models, and then 
 rank the answers.  However, in this case, we want 
 to read a resume from the file resume-text.txt,
 then ask each model to revise the 
 resume.  Then, just like the lab, we ask AI to rank
 the models in the end.


ed-donner:
Great - thanks so much!

PR #1308, opened 2026-04-17
Add 1_foundations lab 1 solution (cleared outputs)

ZoeXu00:
add foundation lab 1 solution


ed-donner:
Great - thanks so much!

PR #1312, opened 2026-04-18
Add deep research multi-agent implementation with autonomous orchestr…

kevinvbc:
Add deep research multi-agent implementation with autonomous orchestration after first negative research evaluation. Clarifying questions for fine-tuning the query, taking into account previous answer. 
An agent decides whether to rewrite the report, do more searches and whether the report is good enough.


ed-donner:
Great - thanks so much!

PR #1326, opened 2026-04-27
Add ginzo_exercises: lab3 with 4 competing sales agents, structured object

soflatechie:
Added code to original lab3 to do the following:

1.  Add a 4th model
2.  Change the original guardrail to one that checks for an off topic prompt
3.  Added a guardrail to check for an unsubscribe line at the end of the email (this was especially difficult)
4.  Added a class for a structured output for the email that included a required property for the unsubscribe line,  this made the guardrail a bit easier to implement.


ed-donner:
Great - thanks so much!

PR #1337, opened 2026-05-02
my progress_contribution

David-gbenga:
## Summary

initiated an open ai api , sent prompts as request messages and got a response 
solved the assignment given in task 1


ed-donner:
Great - thanks so much!

PR #1338, opened 2026-05-03
Execute answers and ranking of answers concurrently.

sirish-darbha:
Request answers from multiple LLMs simultaneously instead of sequentially. Wait until all answers are gathered.
Use the same abstraction to request ranking of the answers from multiple LLMs concurrently. Gather rankings from all LLMs and calculate a cummulative rank.


ed-donner:
Great - thanks so much!

PR #1340, opened 2026-05-04
Prasad gaitonde/6 mcp lab3/h fmcp server

BiprofessionalPrasad:
Day6 MCP Lab3 	3) Managed/hosted remotely running MCP server.
Using Huggingface HF server MCP.


ed-donner:
would it be possible to remove the uv.lock to reduce the size? Thanks so much

ed-donner:
Great - thanks so much!

PR #1343, opened 2026-05-06
Fix MCP server creation issue on Windows + Jupyter by monkey patching "stdio_client" to handle stderr correctly

Carbaz:
# Fix MCP stdio client on Windows + Jupyter

Running MCP servers in Jupyter notebooks on Windows fails with `UnsupportedOperation: fileno` because Jupyter replaces `sys.stderr` with `ipykernel.iostream.OutStream`, a Python object that cannot provide a real OS file descriptor.
When `subprocess.Popen()` tries to use it, it crashes.

## Root cause

Tracked in [python-sdk#1103](https://github.com/modelcontextprotocol/python-sdk/issues/1103).
The MCP client passes Jupyter's captured `stderr` to `subprocess.Popen()`, which internally calls `.fileno()` to get the OS handle, an operation `OutStream` doesn't support.

## Fix proposed in the issue

the [fix to python-sdk#1103](https://github.com/modelcontextprotocol/python-sdk/issues/1103) Consists on passing `stderr=None` when calling `stdio_client()`, so the subprocess uses its own default `stderr` instead of Jupyter's captured stream:

```python
stdio_client(server_params, stderr=None)
```

## Fix applied here

Since `stdio_client()` is called internally by `mcp_server_tools()` and we can't pass parameters to it directly, we monkey patch `stdio.stdio_client` to always pass `stderr=errors_file` before the MCP session is created.

With `None`, the subprocess will bypasses Jupyter's captured stream entirely but error logs will be lost, so we pass an already opened file so we can track it to see outputs.

```python
import mcp

mcp.client.stdio._original_stdio_client = mcp.client.stdio.stdio_client
_errlog = open("mcp_stderr.log", "w", buffering=1)
patched = lambda server, *_: mcp.client.stdio._original_stdio_client(server, _errlog)
mcp.stdio_client = mcp.client.stdio.stdio_client = patched
```

### References

- Issue: https://github.com/modelcontextprotocol/python-sdk/issues/1103
- Fix explanation: https://github.com/modelcontextprotocol/python-sdk/issues/1103#issuecomment-3470416291


Carbaz:
Updated patch to cover imports of `stdio_client` from `mcp` directly, now `from mcp import stdio_client` should be patched too.

This makes the patch to work with week 6 notebooks too, the `from agents.mcp import MCPServerStdio`, wich internally imports from `mcp` and not from `mcp.client.stdio`, will work now.

Carbaz:
> [!WARNING]
> **Issue: Execution failure when paths contain spaces**
>
> The `MCPServerStdio` implementation (and its underlying `mcp` dependency) does not correctly escape spaces in the `command` field.
>
> This is a frequent issue when using `npx` (often located in `...\Program Files\nodejs\`) or `uvx` if your Python environment path contains spaces.
>
> **The Workaround:**
> Use `cmd` as the primary command and move the original command into the `args` list, preceded by the `/c` flag.
>
> * **Instead of:**
>   `playwright_params = {"command": "npx", "args": ["@playwright/mcp@latest"]}`
>
> * **Use:**
>   `playwright_params = {"command": "cmd", "args": ["/c", "npx", "@playwright/mcp@latest"]}`

**P.S.**
Beware of asking Claude about this; you might find yourself on a crusade against `Jupyter`, `Asyncio`, and `_WindowsSelectorEventLoop`. You could end up patching the way the  `mcp` library creates subprocesses only to discover, after a massive refactor, that the root cause was simply the command breaking on spaces. I’ve been there, and it is not pleasant.

Interestingly, a simple Google AI-mode search (even without the full Gemini) flagged space-escaping as a top candidate for the error.

This is a classic "hammer sees only nails" scenario: since Claude is built to handle complex tasks, it ends up imagining massive, intricate problems where they simply don't exist, crafting elaborate fixes for ghosts.

image

ed-donner:
Oh wow this is pretty important stuff, but would you be OK to put it in community_contributions? I'd rather not have it in the main codebase as I'd have to test it extensively first.. thanks so much

Carbaz:
> Oh wow this is pretty important stuff, but would you be OK to put it in community_contributions? I'd rather not have it in the main codebase as I'd have to test it extensively first.. thanks so much

Sure, done. 

Moved to: `5_autogen/community_contributions/Carbaz/fix-mcp-server-on-win+jupyter/2_lab2_autogen_agentchat.ipynb`

Along a `.py`. file with just the fix code to ease copying it to other notebooks.

ed-donner:
Great - thanks so much!

PR #1346, opened 2026-05-07
day1_lab using ollama

arpitt-bharti:


ed-donner:
Great - thanks so much!

PR #1359, opened 2026-05-14
Added a Remote MCP server proxy example and Dockerization

Carbaz:
This PR adds a Jupyter notebook and a Dockerfile that demonstrate how to expose a STDIO-based MCP server as a remote SSE endpoint and connect to it from Python code, covering a scenario not addressed by existing material: shared, remotely reachable MCP servers.

The demo uses `mcp-proxy` to wrap `mcp-server-fetch` (a STDIO-only server) and expose it over SSE, so any number of clients can connect to the same running instance via a standard HTTP endpoint.

The notebook walks through two deployment modes, running `mcp-proxy` directly on the host and running it inside a Docker container, then connects to each using `MCPServerSse` from the OpenAI Agents SDK and lists the available tools.

The Dockerfile packages the whole thing into a single portable image that can be deployed on any OS, machine, or cloud environment, and started and stopped without managing detached processes or PIDs.

Beyond the demo itself, this opens up a practical architectural pattern: In a multi-agent system, instead of each agent spawning its own local MCP subprocess, they can all connect to a single shared MCP service.

This distributes access, avoids duplicating server state, and in cases where the MCP server maintains memory or context, ensures all agents in the system see the same state. 

Deploying via Docker also adds a layer of isolation that makes it safer to expose potentially risky tooling (such as a REPL) inside the service boundary.

The code handles Windows and Unix (Linux/macOS) differences for the local flow. Docker is cross-platform by nature. 

> [!NOTE]
> Everything has been verified on Windows.
>
> Feedback from Linux/macOS users on the local flow would be welcome.

---

**What the notebook teaches**
* How to launch `mcp-proxy` wrapping a STDIO MCP server and expose it as an SSE endpoint.
* How to verify the endpoint is reachable before connecting.
* How to connect from Python using `MCPServerSse` (OpenAI Agents SDK) and list available tools.
* How to stop the proxy cleanly (local PID lookup and kill, both Windows and Unix).
* How to build and run the same setup inside Docker, connect to the containerized endpoint, and stop the container.

**Direct benefits**
* Wrap any STDIO server, from a public repo, a package, or built yourself, and it becomes a **Remote MCP SSE server**.
* Single shared MCP server instance across multiple agents, with shared state and load distribution.
* No per-agent STDIO subprocesses; clients connect over HTTP to the SSE endpoint.
* Docker deployment: reproducible, isolated, easier to operate than detached local processes.
* Network-transparent: the SSE endpoint can be on a remote host, enabling true microservices architectures.
* Safer runtime for sensitive tools (such as a REPL) thanks to container isolation.

**Files added**
* `proxy_fetch_demo.ipynb`: Step-by-step notebook covering the full local and Docker demo flows.
* `Dockerfile`: Minimal image that installs `uv`/`uvx` and runs `mcp-proxy` wrapping `mcp-server-fetch`, on internal port 8000.

### DIFF
--- a/6_mcp/community_contributions/Carbaz/Remote mcp-proxy example/Dockerfile
+++ b/6_mcp/community_contributions/Carbaz/Remote mcp-proxy example/Dockerfile
@@ -1,0 +1,7 @@
+FROM python:3.13-slim
+
+# Install uv (includes uvx)
+RUN pip install uv
+
+# Default command
+CMD ["uvx", "mcp-proxy", "--port=8000", "--host=0.0.0.0", "uvx", "mcp-server-fetch"]
--- a/6_mcp/community_contributions/Carbaz/Remote mcp-proxy example/proxy_fetch_demo.ipynb
+++ b/6_mcp/community_contributions/Carbaz/Remote mcp-proxy example/proxy_fetch_demo.ipynb
@@ -1,0 +1,438 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "3068126d",
+   "metadata": {},
+   "source": [
+    "# Serve an STDIO MCP server over SSE with mcp-proxy\n",
+    "\n",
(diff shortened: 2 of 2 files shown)


ed-donner:
Great - thanks so much!

PR #1364, opened 2026-05-17
Priya's lab3 solution for 2_openai(cleared outputs)

Priya-raja:
Implemented guardrails for email tool in sales agents tasks.


Priya-raja:
@ed-donner Hi Ed. Pls look into the tool guardrails. I found from OpenAI docs to implement it on the email tool.

ed-donner:
Great - thanks so much!

PR #1368, opened 2026-05-17
Add Quinten's lab 3 solution via OpenRouter (outputs cleared)

quintenkocian:


ed-donner:
Great - thanks so much!

PR #1370, opened 2026-05-18
Misi week 5 excercise

mihaly-hazag:
## Summary

This contribution extends the baseline AutoGen dynamic-agent demo with evolving Creator agents. In addition to generating `Agent` modules at runtime, a `Creator` can now generate new `Creator` versions, and `world.py` plans probabilistic Creator switch points before running Agent creation.

## Changes

- Added Creator self-replication support via generated `creatorN.py` modules.
- Added planned Creator switching with `CHANCES_THAT_I_CHANGE_CREATOR`.
- Pipelined execution so each Creator’s planned Agent batch starts while the next Creator is being generated.
- Replaced file-glob recipient discovery with an in-memory registry of actually registered Agents.
- Prevented Agent refinement loops by only allowing peer refinement for initial idea requests.
- Added model-name protection so generated code keeps `OpenAIChatCompletionClient` on `gpt-4o-mini`.
- Made generated output paths relative to the contribution folder.


ed-donner:
Great - thanks so much!

PR #1373, opened 2026-05-18
openllmetry instrumentation

jason-godboldsmith:
openllmetry instrumentation with Dynatrace backend


jason-godboldsmith:
openllmetry instrumentation with Dynatrace backend (backend can be easily swapped by changing the TRACELOOP environment variables in .env file - example .env file provided)

ed-donner:
Great - thanks so much!

PR #1376, opened 2026-05-21
Add Quinten's Week1 Lab4 Solution (outputs cleared) with Digital Twin Code

quintenkocian:
Added headshot to Digital Twin project (included in /me directory)

Made a mistake when committing - first commit didn't actually have outputs cleared. Both commits have the wrong lab identifier. These are both related to Week1 Lab4


ed-donner:
Ahh this is great - but could you remove the image from the PR, or make it a bit smaller than 1MB? Thanks so much

quintenkocian:
Hey @ed-donner I think I figured out the Git side of this - but you should now see the headshot file size improved (reduced) by about an OOM down to ~100 kB. Does that work?

ed-donner:
Great - thanks so much!

PR #1395, opened 2026-06-04
correct "f-string without any placeholders" error in 3_lab3.ipynb

AAYUSH-GUPTA-coder:


ed-donner:
Great - thanks so much!

PR #1397, opened 2026-06-05
Fix MCP Jupyter fileno workaround notes for Windows

igoradriano:
This PR adds a small but important reliability fix for Week 6 MCP labs running in Jupyter notebooks on Windows.

What was changed:

Added a stdout/stderr reset cell before MCP server startup in:

6_mcp/1_lab1.ipynb
6_mcp/2_lab2.ipynb
6_mcp/3_lab3.ipynb

Added an English note in each notebook explaining:
why this workaround is needed (fileno-related startup issues in Jupyter),
and that users should reselect the notebook kernel and rerun from the top after adding/updating the reset cell.
Why:
In some Windows + Jupyter environments, MCP server startup can fail due to stream handle behavior (fileno issues). Rebinding sys.stdout and sys.stderr to sys.__stdout__ / sys.__stderr__ before creating MCP servers avoids this failure mode.

User impact:
Learners are less likely to hit MCP startup errors in Week 6 notebooks, and the notebook guidance now clearly documents the required kernel re-selection step.


ed-donner:
Hey..
The official recommended workaround is to use WSL instead...
If you'd like to add this change as the fallback for those that stick with Windows, then please could you add it as a Markdown file in community_contributions instead, and I'll link to it.. I'd rather not update the main labs, especially given the WSL recommendation.
Did everything work for you on Windows? I hear this is much more stable these days.. I'd be super interested to hear if it worked throughout week 6.
Ed

PR #1398, opened 2026-06-05
Replaced Big If statement by tools dictionary

rafahsolis:
This is a way to make tool definition more readable.
Thanks for your course!


ed-donner:
Yes, that's the better approach, and we do something similar coming up in the course. But any changes like this need to be in community_contributions please..

rafahsolis:
Sorry!
Will do! thanks for your course Ed, I'm enjoying it a lot and will do more of your courses!

ed-donner:
fab! thank you! look forward to welcoming you on many more..

PR #1405, opened 2026-06-11
Add sofcal1 lab1 solution (cleared outputs)

sofcal1:
- Created a deutsch to english translator


ed-donner:
Great - thanks so much!

PR #1409, opened 2026-06-13
Lab 2 | Community Contributions | Orchestrator Worker Pattern to solve the trolley problem 

kuriangeorgebinu:
+ Added pattern using the orchestrator worker pattern which was different from the parallel pattern that was in the lab 2 lecture.
+ The lecture asked each agent from code with an evaluator at the end.  
+ Instead, I added an orchestrator with the task to make the decision to know which agents to call and then added a judge at the end to rank the answers in terms of ethicality, logical thinking and reasoning. (Based on the results claude is the winner of all three categories).


ed-donner:
please clear outputs.. many thanks

kuriangeorgebinu:
@ed-donner cleared the output. please check if this satisfies.  Appreciate your feedback.

ed-donner:
Great - thanks so much!

PR #1415, opened 2026-06-19
deep-research-rwk-orchestrator-agent

rwkopcke:
created an interactive agent to clarify the query as necessary
created an orchestrator agent to guide the workflow (the workflow seemed clear and mandatory, so allowing the orchestrator to choose the use and sequence of agents was not an option)
eliminated the python code loop in the search functions
created a search filter agent
created a report review agent
created an agent to publish the report


ed-donner:
Great - thanks so much!

PR #1422, opened 2026-06-27
Deep research Agent with clarify questions and anwser on UI

dragonlong206:
I just modify your deep_research tool to make it auto ask clarification question to user. After each question, the agent pause, user must answer the question and then the agent decide to continue clarifying or plan the search. Progress is updated on UI.


ed-donner:
Great - thanks so much!

PR #1434, opened 2026-07-02
Add bhargav's lab 1 solution (cleared outputs)

bhargavndesai25:
The notebook contains the lab 1 solution with cleared outputs


ed-donner:
Great - thanks so much!

PR #1445, opened 2026-07-05
MGM - Week 5 - Agent Frameworks  Additional Labs

mgmarques:
The code in these labs executes the same exercises on each agent framework from Week 5.

The intention is a simple example of a multi-step orchestration pipeline, based on a local LLM (Ollama running Qwen 2.5). It should be used only for learning purposes and comparison of different frameworks, especially their specific aspects of observability implementation and an automated workflow engine with human-in-the-loop approval.

The goal is to receive an initial product request, dynamically generate a marketing slogan, translate it into Spanish, pause execution for a human confirmation step, and conditionally write the final output to the local file system using the Model Context Protocol (MCP).


ed-donner:
Hi there,
This is a PR that is more than 1 million lines of code, and more than 3,000 files changed. It's too large for Github to even display it.
It would probably take me several years to review this. Neither of us would benefit from that.
It should be entirely possible to do exactly what you say, with all agent frameworks, in a small project of around 1,000 lines.
I'm closing this PR - please only submit PR with a maximum of 2,000 lines of code and 10 files.
Thanks much
Ed

mgmarques:
Okay, Ed, 99% of that is Mastra's node stuff.

PR #1446, opened 2026-07-05
Mastra cleanup

mgmarques:
The code in these labs executes the same exercises on each agent framework from Week 5.

The intention is a simple example of a multi-step orchestration pipeline, based on a local LLM (Ollama running Qwen 2.5). It should be used only for learning purposes and comparison of different frameworks, especially their specific aspects of observability implementation and an automated workflow engine with human-in-the-loop approval.

The goal is to receive an initial product request, dynamically generate a marketing slogan, translate it into Spanish, pause execution for a human confirmation step, and conditionally write the final output to the local file system using the Model Context Protocol (MCP).


ed-donner:
This is amazing, but I'm not going to have the capacity to review a 13-14,000 lines of code PR.. would it be possible to just submit a markdown file that describes the project and link to your repo instead? Thanks so much

PR #1451, opened 2026-07-08
Added Class Action Agents

eterry28:
This crew will go out and research open class action suits.


eterry28:
I fixed a problem with my crew project. It was added as a submodule last time.

ed-donner:
Great - thanks so much!

PR #1452, opened 2026-07-09
Add fix description for Windows path handling issue on MCP commands.

Carbaz:
### Summary
Add a Windows troubleshooting note for MCP commands launched through a command whose resolved executable path can contain spaces, and recommend wrapping the command with `powershell` or `cmd`.

### What changed
- Added guidance explaining that MCP may expand a command into a path like `C:\Program Files\...`
- Recommended using a native Windows shell interpreter to avoid path issues
- Included an example replacement:
  - from `{"command": "npx", "args": ["@playwright/mcp@latest"]}`
  - to `{"command": "powershell", "args": ["/c", "npx", "@playwright/mcp@latest"]}`
- Clarified that the original command and its arguments should be passed in `args` prefixed by `"/c"`

### Why
This prevents Windows notebook launches from failing when the resolved command path contains spaces, and keeps MCP server startup compatible with Windows shell path handling.


ed-donner:
Thanks so much for sorting this

PR #1453, opened 2026-07-09
Add Leckan's lab 1 solution (cleared outputs)

Leckan:
Leckan's solution to Lab 1 added.


ed-donner:
Great - thanks so much!

PR #1458, opened 2026-07-12
Omniroute

waqarwld:
this branch will implement local llm like omniroute or ollama .


waqarwld:
this just keep separate branch not for merging into main branch

PR #1463, opened 2026-07-14
Add routing pattern version of lab 2

qqjjz12:
Adds a community contribution notebook that rewrites Lab 2 using Anthropic's routing workflow pattern. The notebook routes the generated question to a specialist model and audits the route choice with Grok via OpenRouter.


ed-donner:
Great - thanks so much!

PR #1464, opened 2026-07-15
add community contributions for llm as a judge

jsjasee:
Added 2 new features - a guardrail that checks the users question and LLM as a judge to check the output of the LLM.


ed-donner:
Great - thanks so much!

PR #1471, opened 2026-07-16
Add Shru's lab 1 solution (cleared outputs) and it also contains all …

skyshru:
…the documentation for day 1 so anyone can reproduce it


ed-donner:
Great - thanks so much!

PR #1487, opened 2026-07-22
This is the first lab on agents

Onyekaeveristus:
It builds a new prompt that shows all 3 options back to the AI and asks: "Which one of these is the most complex/challenging — the kind of thing that really needs a multi-step AI agent, not just a quick one-off task? Just reply with the number."
It sends that prompt and gets back a short reply, hopefully just "1", "2", or "3".
It uses re.search to pull the first digit (1–3) out of that reply, in case the AI adds extra words around it.
If no number is found for some reason, it safely defaults to Option 1.
It then looks up that chosen option's full text and prints it


ed-donner:
Hey,
I'm super grateful that you've worked on this and are sharing your contribution. It looks like this might not be ready to be merged - please can I ask you to review this:

https://edwarddonner.com/pr

The most common issue is that you've included a file outside community-contributions. This would impact all students on the course. You can see this by going to the "Files changed" tab in this PR; the only changes should be to files in community-contributions.

If there's a good reason to change files outside community-contributions, like an important bug fix, then please let me know (and thank you!)

The 2nd most common reason is that you've not cleared outputs of Notebooks, leading to large code changes. Instructions at the link above.

The 3rd common reason is that you've included too many changes, such as many thousands of lines of code, or a large number of files. I'm grateful for your work, but see the link above for ways we can include this without the repo getting too large.

The PR details in this Github screen shows you the files changed, and at the top right the number of lines added (in Green) and deleted (in Red) by this PR. Typically the Green number should be less than 2,000 and the red number should be 0.

I very much look forward to merging when this is addressed. Thanks so much
Ed

PR #1488, opened 2026-07-23
fix: work around Windows Jupyter stderr incompatibility

smanefield:
When running under IPykernel (e.g. Jupyter Notebook/Lab or VS Code notebooks) on Windows, sys.stderr is not backed by a real file descriptor and fileno() raises io.UnsupportedOperation. 

Passing subprocess.DEVNULL avoids this incompatibility.
It is minimally invasive, leaving the behavior unchanged everywhere else.
It is designed to not affect Non-Windows users and/or users who are not running from a Jupyter notebook.

### DIFF
--- a/6_mcp/backend/accounts_client.py
+++ b/6_mcp/backend/accounts_client.py
@@ -1,7 +1,29 @@
+import sys
 import mcp
 from pathlib import Path
 from mcp.client.stdio import stdio_client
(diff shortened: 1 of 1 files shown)


ed-donner:
massive thanks - and for the defensive checks there. Hugely appreciated.

PR #1489, opened 2026-07-23
fix:  allow the latest playwright to browse local files

smanefield:
Playwright security seems to have been recently hardened to restrict the browsing of any local files.
It makes sense.
In our case though, we point playwright's MCP server at specific local files like game.html, which we know to be safe.
We can overcome Playwright's reluctance to browse locally by adding a single parameter to the args list for the MCP Server: 
`"--allow-unrestricted-file-access"`


ed-donner:
And again - superb. Thank you so much.

PR #1517, opened 2026-07-31
Community contributions w3d5

jsjasee:
Summary:
1. Changed crewai crew from sequential to hierarchical
2. Experimented with openrouter models which are cheaper
3. Added a manager agent so that the engineering-lead agent can only focus on the system design and not balance the task of delegating tasks.

Result:
1. Generated a comparable final product for only 1.50 dollars, which shows that open source models can also have decent performance.


ed-donner:
Great - thanks so much!

PR #1534, opened 2026-08-09
add community contributions for week 4 day 3 based on challenges

jsjasee:
Completed challenges for w4d3:
1. Added memory to the browser agent and confirmed that it remembers its output by prompting it to recall the name of the first news title
2. Confirmed that the middleware works by using different websites and also created a block-list and made the middleware async


ed-donner:
Great - thanks so much!

PR #1538, opened 2026-08-11
Fix: Capture stderr in sandbox_python tool and make agent names match…

tony-howard-fsu:
…able

Two related bugs in the engineering_team reference implementation:

1. run_sandbox_python discarded stderr: Python's unittest module and uncaught exceptions write to stderr by default, not stdout. The tool was silently throwing away the only output that explained test failures or crashes, leaving agents unable to distinguish success from failure. Now returns exit code, stdout, and stderr clearly labeled.

2. Agent role names were unmatchable: When using hierarchical processes, delegate_to_coworker calls failed to find agents because roles used long descriptive names instead of concise titles. Updated all agent roles to match their names: backend_engineer, frontend_engineer, etc. Preserved the original role descriptions as the first line of each goal for clarity.

These fixes prevent wasted LLM calls from failed agent delegation and allow test failures to be visible to the test engineer agent.

### DIFF
--- a/3_crewai/reference/engineering_team/src/engineering_team/config/agents.yaml
+++ b/3_crewai/reference/engineering_team/src/engineering_team/config/agents.yaml
@@ -1,7 +1,8 @@
 engineering_lead:
   role: >
-    Engineering Lead for the engineering team, directing the work of the engineers
+    Engineering Lead
   goal: >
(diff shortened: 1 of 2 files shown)


ed-donner:
That's terrific - huge thanks

PR #1541, opened 2026-08-11
Exercises for the Foundation Week!

supezlopez:
Foundation Week Exercises


ed-donner:
looks like one of the outputs needs to be cleared please.. also usually better not to include the PDF in the repo if poss

ed-donner:
Great - thanks so much!

PR #1543, opened 2026-08-12
add week 4 day 4 contributions

jsjasee:
Updated the prompt to accept playwright MCP server tools and did a few testing to ensure the tools are running smoothly. Also limited the playwright tools the agent has access to, to ensure the agent does not get confused with the context and cause chaos.


ed-donner:
Great - thanks so much!

PR #1544, opened 2026-08-12
Add community contribution: week 1 day 1 lab1 solution

sevarudakov:
Solution to the week 1 day 1 exercise: chained LLM calls that pick a business
area, identify a pain-point in it, then propose an Agentic AI solution.

The twist is that each call returns JSON and the next prompt reads named fields
out of it, rather than piping raw prose forward - which breaks as soon as a model
answers conversationally. There's also a final call that checks the idea against
what already exists, since a chain of agreeable models will talk itself into
anything.

Adds one file under 1_foundations/community_contributions. Notebook outputs are
cleared.

### DIFF
--- a/1_foundations/community_contributions/1_lab1_seva.ipynb
+++ b/1_foundations/community_contributions/1_lab1_seva.ipynb
@@ -1,0 +1,525 @@
+{
+ "cells": [
+  {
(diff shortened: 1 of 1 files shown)


ed-donner:
Great - thanks so much!

PR #1567, opened 2026-08-21
langchain agent orchestration example

rishabhToshniwal:
Week 4 community contribution: a small `create_agent` example that orchestrates a main agent, tools, structured output and a nested apparel LLM subagent, with `MemorySaver` so follow-up questions keep the city in context.


ed-donner:
Ahh we were speaking in Udemy right? Very nice, thank you for contributing this
Ed

PR #1570, opened 2026-08-23
add w6d3 community contributions firecrawl mcp server

jsjasee:
Added 2 versions of firecrawl mcp server - Stdio or Streamable HTTP


ed-donner:
Great - thanks so much!

PR #1571, opened 2026-08-23
AI Assistant for courses lab 3

Abhishek-Maury:
Implemented an AI Assistant for courses using Gemini, an email recording tool, and a second LLM to evaluate whether the first LLM's response is related to AI/ML business queries. Thank for your guidance on every step😊 @ed-donner


ed-donner:
You bet! Thanks for the contribution..

PR #1577, opened 2026-08-26
Exercise 1: LLM-as-a-Judge guardrail using Gemini (gemini-3.5-flash-lite)

torisgup:
## Summary
Added Exercise 1 solution to 3_lab3.ipynb — implements an LLM-as-a-Judge guardrail pattern.

## What was added
- `is_professional(reply)` function: uses a second Gemini LLM call to evaluate whether the chatbot's reply is strictly professional/work-related
- `chat_with_guardrail(message, history)` function: the full agent loop with tool calling + guardrail check
- Gradio UI launch for the guardrail chatbot

## How it works
1. User sends a message
2. Main LLM (gemini-3.5-flash-lite) generates a reply (with email tool support)
3. Judge LLM evaluates the reply — responds YES or NO
4. If YES → reply is returned to user
5. If NO → user is redirected back to professional topics

## Notes
- Uses Google Gemini API (free tier) via OpenAI-compatible endpoint
- Model: `gemini-3.5-flash-lite`
- No OpenAI API key required

### DIFF
--- a/1_foundations/3_lab3.ipynb
+++ b/1_foundations/3_lab3.ipynb
@@ -40,68 +40,167 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 2,
    "metadata": {},
    "outputs": [],
    "source": [
     "# If you don't know what any of these packages do - you can always ask ChatGPT for a guide!\n",
     "\n",
+    "import os\n",
     "from dotenv import load_dotenv\n",
     "from openai import OpenAI\n",
     "from pypdf import PdfReader\n",
     "from IPython.display import Markdown, display\n",
     "import gradio as gr\n",
-    "import json"
+    "import json\n"
    ]
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 3,
    "metadata": {},
    "outputs": [],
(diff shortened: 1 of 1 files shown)


ed-donner:
Hey,
I'm super grateful that you've worked on this and are sharing your contribution. It looks like this might not be ready to be merged - please can I ask you to review this:

https://edwarddonner.com/pr

The most common issue is that you've included a file outside community-contributions. This would impact all students on the course. You can see this by going to the "Files changed" tab in this PR; the only changes should be to files in community-contributions.

If there's a good reason to change files outside community-contributions, like an important bug fix, then please let me know (and thank you!)

The 2nd most common reason is that you've not cleared outputs of Notebooks, leading to large code changes. Instructions at the link above.

The 3rd common reason is that you've included too many changes, such as many thousands of lines of code, or a large number of files. I'm grateful for your work, but see the link above for ways we can include this without the repo getting too large.

The PR details in this Github screen shows you the files changed, and at the top right the number of lines added (in Green) and deleted (in Red) by this PR. Typically the Green number should be less than 2,000 and the red number should be 0.

I very much look forward to merging when this is addressed. Thanks so much
Ed

PR #1581, opened 2026-08-28
Multi-LLM Debate simulator by Waibhav (cleared outputs)

aquagreen1000:
- This notebook simulates a debate among three different LLMs on a topic.
- Topic can be hard coded. E.g. - "Which is the best FIFA team"?
- With small code modification, it can be a system input
 
- Each debater gets 4 turns to provide rebuttal and make their own argument.
- At the end, full transcript is created
- Additionally, each debater provided a score to other debator


ed-donner:
Great - thanks so much!

PR #1591, opened 2026-09-01
My Week 1 Lab 2 community contributions (cleared outputs)

teo2001-repos:
Hi Ed!  Hope you are doing very well! In advance, thank you for your time. 

Comments below also in the .ipynb file for anyone else who would like to understand what I did in this community contribution:

---
My contribution consist of three sets of changes. (Per Python's PEP-8 I hope my code is somewhat clear and readable enough that it does not need heavy commenting, fingers crossed).

Just for the sake of the exercise, I didn't ask any LLM to create the code or come up with the idea, so I could put just some thought to it.

### 1. To the Original lab 

- Import and use the official SDKs for Anthropic, Google, Groq, xAI, OpenRouter, and Ollama, so it can be compared side by side in the same cells.
- I added the shared `record()` helper in the Groq and Ollama cells replacing the `display` / `append` separate calls in those cells (so all the models use the record() function).

### 2. Improved workflow

This section starts after the original lab cells. It follows the lecture prompt from Ed to try to make the workflow more sophisticated. I aimed to use only the content from the course so far and what Ed included in all the GitHub guides (e.g. Python's async), and to keep the workflow, so the essence is still there.

- I created a dict for the LLM providers/models, so parallel runs with async are easier to manage and extra per-model settings and/or more models/providers can be added later.

- Added to the judge prompt XML tags around the question, format, and responses. I read some time ago on Claude’s prompting guidance and related papers on structured prompts that this helps the LLM to understand the structure and in general is a good practice (arXiv papers 2510.22956, 2509.08182).

- Also I added a prompt dict, easier management of prompt templates, and in the case of the judge, it could be filled for the particular call with the `str.format()` method.

- New function helpers (including API-key prefix printing).

- To be faithful to the original lab, `llm_call()` keeps Ed’s `reasoning_effort="none"` behavior for `gpt-5.4-nano` , it works from a conditional.

- The async function llm_call is a starting point: it can be modified to expand/add conditions/features, for example: skip disabled models (e.g. in the LLMs dict with an "disabled" property), pull `max_tokens` / `reasoning_effort` (it could be from the prompts dict), etc, switch SDKs (I coded this new section to use the OpenAISDK. As recreating an improved version of the original lab code.)

### 3. Additional agentic design pattern

- I added an additional agentic workflow design pattern at the end, as an example to start a version 3 of the code of this lab (v1 the original, v2 the improved workflow, v3 adding additional patterns or calls asking new things). The added pattern is **routing**. 

It works this way:

1. Models vote using the agentic workflow parallel design pattern on which model should write the opening question (an LLM council).
2. Votes are counted. On a tie, `max()` keeps the first model that reached that count. (This could be changed but for the moment let's say it's "destiny").
3. That code routes the question-generation prompt to the winner model.
----

Kind Regards,

Theo.

### DIFF
--- a/1_foundations/community_contributions/theo/2_lab2.ipynb
+++ b/1_foundations/community_contributions/theo/2_lab2.ipynb
@@ -1,0 +1,1076 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Welcome to the Second Lab - Week 1, Day 3\n",
+    "\n",
+    "Today we will work with lots of models! This is a way to get comfortable with APIs."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "<table style=\"margin: 0; text-align: left; width:100%\">\n",
+    "    <tr>\n",
+    "        <td style=\"width: 150px; height: 150px; vertical-align: middle;\">\n",
+    "            <img src=\"../../../assets/stop.png\" width=\"150\" height=\"150\" style=\"display: block;\" />\n",
+    "        </td>\n",
+    "        <td>\n",
+    "            <h2 style=\"color:#ff7800;\">Important point - please read</h2>\n",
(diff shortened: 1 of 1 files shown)


ed-donner:
Great - thanks so much!

PR #1593, opened 2026-09-02
Ashelman_David Updated mcp_servers.py

Dashelman:
Pinning MCP to version one in the MASSIVE Server - causes all MCP Servers to communicate with each other


ed-donner:
Ahh you've included in here the .cmd change, but that's a windows-only fix (and usually not required - just an occasional problem). Would it be possible to try this without that change? Many thanks

Dashelman (commented):
Removed .cmd from MCPServerStdio's tools

Dashelman (commented):
updated to not use .cmd

ed-donner:
thank you David!

PR #1594, opened 2026-09-02
Ashelman_David-Update sandbox_tools.py

Dashelman:
Sandbox_tools.py in Projects\agents\3_crewai\coursework\engineering_team\src\engineering_team\tools

Edited line 69 to NOT error when checkmarks are written to file by adding encoding="utf-8" to the path.write_text() call


ed-donner:
Great - thanks so much!

PR #1597, opened 2026-09-03
Create 4_lab4.ipynb_Ashelman

Dashelman:
Added functions to verify which MCP_servers are connected, which version of MCP the notebook is using


ed-donner:
Great - thanks so much!

PR #1598, opened 2026-09-04
Create api.py - New & improved - Ashelman_David - addition of 2 decorators 

Dashelman:
# These errors occur w/o the next two (2)@app.get() calls
# INFO:     127.0.0.1:56973 - "GET / HTTP/1.1" 404 Not Found
# INFO:     127.0.0.1:49266 - "GET /json/version HTTP/1.1" 404 Not Found

@app.get("/")
def root():
    return {"status": "ok"}

@app.get("/json/version")
def version():
    return {"version": "1.0"}


ed-donner:
Great - thanks so much!

PR #1609, opened 2026-09-09
Middleware to diagnose browser calls in playwright

d215steinberg:
Add this middleware to an agent using the Playwright MCP tools.  The middleware will

1. Log all URLs passed to **browser_evaluate**
2. Log all errors


ed-donner:
Great - thanks so much!

PR #1620, opened 2026-09-14
Resolve rank/position queries by structural order on page, not by tex…

d215steinberg:
The search for the top-ranked Hacker News article was actually returning the #3 article.  The results were being accepted because the success criteria specified only "a specific article on the Hacker News front page."  The reason for the failure was that the agent was searching for /1\./.  Although the top article is indeed marked with "1.", this label is generated by the browser, whereas the "snapshot" is taken from the DOM.

I have updated the WORKER_PROMPT with specific instructions regarding position/rank queries.  I have also updated the success criteria to instead upon the top-ranked article.


ed-donner:
Great - thanks so much!

